PR #90, opened 2024-04-24
Feature/asset 81 restrict access to groups

### LINKED ISSUE #81: Authentication: Restrict access to Users with the "asset.swissgeol" group
When Users authenticate using eIAM via Cognito, they first have to be assigned to the group "asset.swissgeol" in order to be authorized to access the Assets application.

Currently, any authenticated User gets access with the Viewer role.

Restrict the application access to users with the group "asset.swissgeol".

The group "asset.swissgeol" is already available on DEV but not yet on PROD and INT. Someone with permission has to add them.

Open Questions
1. What should be shown when a User is authenticated but not authorized for "asset.swissgeol"? How does Boreholes do this?

**Acceptance criteria**
1. Users only can access Assets when they are in the "asset.swissgeol" group
2. The user group exists on INT and PROD

TIL-EBP:
resolves #81 

Restricts access to the application via user groups
Adds a simple error page when a request fails

Add the respective groups to your .env.local to make it work


apps/server-asset-sg/src/app/jwt/jwt-middleware.ts
```diff
@@ -81,6 +82,16 @@ export class JwtMiddleware implements NestMiddleware {
         );
     }
 
+  private isAuthorizedByGroupE(jwt: Jwt): E.Either<Error, Jwt> {
+      const authorizedGroups = process.env.OAUTH_AUTHORIZED_GROUPS?.split(',');
```
daniel-va, line 86: - Trim each group so that checks don't fail when we have spaces between groups in our config.
- Lowercase each group so that checks don't fail when we use different casing in our config.

apps/server-asset-sg/src/app/jwt/jwt-middleware.ts
```diff
@@ -81,6 +82,16 @@ export class JwtMiddleware implements NestMiddleware {
...
+      if (!authorizedGroups) {
+        return E.left(new Error('An internal server error occurred. Please try again.'));
+      }
+      const userGroups = (jwt.payload as JwtPayload)['cognito:groups'] as string[];
```
daniel-va, line 90: Similar to `authorizedGroups`: lowercase these groups for a case-insensitive comparison.
I assume that trimming is not be necessary here.

libs/auth/src/lib/services/auth.interceptor.ts
```diff
@@ -22,11 +24,16 @@ export class AuthInterceptor implements HttpInterceptor {
                 redirect_uri: window.location.origin,
                 response_type: this._oauthService.responseType,
             });
-            return throwError(new HttpErrorResponse({ status: 401, statusText: 'Unauthorized' }));
-        } else if (!token) {
-            return throwError(new HttpErrorResponse({ status: 401, statusText: 'No Token' }));
+            return EMPTY;
+        }  else if (!token) {
+          return EMPTY;
```
daniel-va, line 29: Is there no logout required when no token is present? Does the redirect to the login form still happen?

daniel-va (changes requested):

libs/auth/src/lib/services/auth.interceptor.ts
TIL-EBP: I remembered the logic we were using there: 
If I have a token in the storage but it is invalid, I shoud be logged out
If I dont have a token, then the login process has not yet finished (oauth2 flow)
If i have a token and it is valid (the else-case), then we redirect to error page when an error occurs

So if I added the logout in the !token case, then there would be an indefinite redirect to the login page.

apps/server-asset-sg/src/app/jwt/jwt-middleware.ts
TIL-EBP: added both the trim and toLowerCase just in case

libs/auth/src/lib/services/auth.interceptor.ts
daniel-va: Alright :)

daniel-va (approved):

PR #114, opened 2024-05-16
Feature 106: Bug - Fehlerseite bei Suche

### LINKED ISSUE #106: Bug: Fehlerseite bei Suche nach sgsid:xxx
**Beschreibung**
Wenn man spezifisch nach einer SGS-Id sucht mit dem Suchinput z.B. sgsid:1703, dann taucht eine Fehlerseite auf.

1. Gehe auf assets
2. Schreibe "sgsid:1703" in das Suchfeld
3. Eine Fehlerseite erscheint

**Beobachtetes Verhalten**
Eine Fehlerseite erscheint.

**Erwartetes Verhalten**
Die Karte bleibt bestehen. Die Ergebnisse werden wie gewohnt angezeigt.
Auch wenn der Backend Call einen Fehler meldet, sollte an diesem Ort nicht auf die Fehlerseite weitergeleitet werden. Das ist nicht benutzerfreundlich.

**Screenshots**
image
image

daniel-va:
Resolves #106.

Replaces the error page at `/{lang}/error` with overlay alerts.

The search described in #106 is already solved in #102.
It will go live in the search rewrite of #99.


libs/client-shared/src/lib/features/alert/alert-list/alert-list.component.scss
TIL-EBP, line 1: Did you talk to Simon about the design of the alert? also the alert component itself, the icons and animations etc.?
I think before we spend to much time designing, we should check with simon about this, as he also does the design for boreholes, and they should look the same (if they already do, nevermind what i said)

libs/client-shared/src/lib/features/alert/alert-list/alert-list.component.ts
```diff
@@ -0,0 +1,18 @@
+import { Component, inject } from '@angular/core';
+import { Store } from '@ngrx/store';
+
+import { AlertId } from '../alert.model';
+import { AlertEntry, AlertState } from '../alert.reducer';
+import { selectAlerts } from '../alert.selectors';
+
+@Component({
+  selector: 'ul[app-alert-list]',
```
TIL-EBP, line 9: Same question as above: I have never seen this. What is the benefit over simple selectors?

libs/client-shared/src/lib/features/alert/alert/alert.component.scss
```diff
@@ -0,0 +1,118 @@
...
+  gap: 0.75rem;
+  overflow: hidden;
+  cursor: pointer;
+
+  animation: 300ms ease-in slide-in;
+  transform-origin: bottom;
+}
+
+:host.is-removed {
+  animation: 250ms ease-out slide-out;
```
TIL-EBP, line 24: Are these animations in accordance with boreholes and Simons design?

TIL-EBP (commented):
I like to alert popups, but i think we need to discuss with simon about when to show alerts and when to show the error page and also match the design with boreholes

libs/client-shared/src/lib/features/alert/alert.reducer.ts
TIL-EBP, line 1: Wouldn't it make sense to have these files in the state folder? (reducer, slector, actions and model?)

libs/client-shared/src/lib/features/alert/alert-list/alert-list.component.html
```diff
@@ -0,0 +1,7 @@
+<ng-container *ngFor="let entry of alerts$ | async; trackBy: getAlertId">
+  <li
```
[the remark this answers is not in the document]
daniel-va: Just for readability, and to separate the loop from its content. It's not strictly better in any way, although starting from v17 Angular supports [`@for`](https://angular.io/api/core/for) for this exact thing.

libs/client-shared/src/lib/features/alert/alert-list/alert-list.component.ts
daniel-va: When using an element selector (e.g. `app-alert-list`), we essentially create a new, custom HTML element. This makes sense when there is no existing element that matches the semantic meaning of what we're trying to do. However, if there is such an element, using an attribute selector like here allows us to reuse that element. 

This is especially useful in cases like `li`, which should always follow directly after their parent `ul`/`ol`. With attribute selectors, this is possible, e.g. this is valid HTML:
```html
<ul app-alert-list>
  <li app-alert-item>My Alert</li>
</ul>
```
While this is not:
```html
<app-alert-list>
  <ul>
    <app-alert-item>
      <li>My Alert</li>
    </app-alert-item>
  </ul>
</app-alert-list>
```
Finally, attribute selectors also allow us to generate less HTML when we want to wrap a component's content inside a specific HTML element, just like here. Other elements besides `li` where this is very useful is with things like `button`, `dialog`, `footer`, and more.

apps/client-asset-sg/src/app/app.component.html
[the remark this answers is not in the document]
daniel-va: This has not been discussed with Simon yet - as far as I'm aware, Alerts/Errors in General have not (yet) been standardized across our projects. I will cross-check with Simon if there are any plans to do so in the near future. I assume its no problem to use the design included here for now - we can still change it later with little effort.

In regards to the comment above: This PR directly references the request in #106 to not redirect to an error page when a search fails. Even if the search still errors, this PR solves the expected behavior in the issue.

libs/client-shared/src/lib/features/alert/alert.reducer.ts
daniel-va: Personally, due to the size of the project and the fact that we might want to move these files in the near future, I think that grouping everything by feature rather than by layer/type makes things a lot easier for us. We can discuss this personally, as I think it would make sense to "officially" decide which way we want go for for now.

apps/client-asset-sg/src/app/components/error/error.component.html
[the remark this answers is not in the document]
daniel-va: I re-added the error page, but as an overlay instead of a separate page.

TIL-EBP (approved):

PR #143, opened 2024-05-30
Feature 83: Login Flow when Cancelling

### LINKED ISSUE #83: Authentication: Login Flow when cancelling
What should a User see when the login flow using eIAM is cancelled?

Currently, the emtpy Asset application with an infinite loading bar is shown.

Options:
- Automatic redirect to login screen
- Login button
- Access denied screen

daniel-va:
Resolves #83.

Adds a splash screen that is shown instead of the application itself whenever the user has not been authenticated yet, or when the authenticated user does not have the correct permissions.


TIL-EBP (changes requested):
Once you set the Authstate to success, everything is fine

PR #153, opened 2024-06-11
Feature 150: Bug - File Download schlägt fehl

### LINKED ISSUE #150: Bug: File download schlägt fehl
**Beschreibung**
Wenn man ein File herunterladen oder in einem neuen Tab öffnen will, so schlägt das Fehl.
Es scheint keine Authentifizierung vorhanden zu sein.
Siehe Diskussion der Lösungsvarianten hier https://stackoverflow.com/questions/29452031/how-to-handle-file-downloads-with-jwt-based-authentication

1. Asset Detailansicht
2. File Download oder im neuen Fenster öffnen
3. Es erscheint die Fehlermeldung «error: no token found».

**Beobachtetes Verhalten**
Eine Fehlermeldung

**Erwartetes Verhalten**
Das File wird heruntergeladen.

daniel-va:
Resolves #150.

Files are now first downloaded via AJAX and then exposed via ObjectURL. This allows us to authenticate the downloads with our auth token.

### DIFF
--- a/libs/asset-viewer/src/lib/components/asset-search-detail/asset-search-detail.component.ts
+++ b/libs/asset-viewer/src/lib/components/asset-search-detail/asset-search-detail.component.ts
@@ -1,16 +1,44 @@
+import { HttpClient } from '@angular/common/http';
 import { ChangeDetectionStrategy, Component, Input } from '@angular/core';
+import { map } from 'rxjs';
 
 import { ApiError } from '@asset-sg/client-shared';
 import { ORD } from '@asset-sg/core';
+import { AssetFile } from '@asset-sg/shared';
 
 import { AssetDetailVM } from '../../state/asset-viewer.selectors';
 
 @Component({
-    selector: 'asset-sg-asset-search-detail',
-    templateUrl: './asset-search-detail.component.html',
-    styleUrls: ['./asset-search-detail.component.scss'],
-    changeDetection: ChangeDetectionStrategy.OnPush,
+  selector: 'asset-sg-asset-search-detail',
+  templateUrl: './asset-search-detail.component.html',
+  styleUrls: ['./asset-search-detail.component.scss'],
+  changeDetection: ChangeDetectionStrategy.OnPush,
 })
 export class AssetSearchDetailComponent {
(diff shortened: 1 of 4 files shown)


libs/asset-viewer/src/lib/components/asset-search-detail/asset-search-detail.component.ts
```diff
@@ -1,16 +1,44 @@
...
+
+      anchor.setAttribute('style', 'display: none');
+      anchor.href = url;
+      anchor.rel = 'noopener noreferrer';
+      if (isDownload) {
+        anchor.download = file.fileName;
+      } else {
+        anchor.target = '_blank';
+      }
+      document.body.appendChild(anchor);
```
TIL-EBP, line 36: I had a similar task in GB3 just recently, where we had to download a file from a http request with responsType of blob.
WE explicitly decided to not go this way because it bypasses all the security measures put in place by Angular. Instead, we decided to use the FileSaver library for this task: 

[Filesaver](https://www.npmjs.com/package/file-saver)

Here are two articles talking about this in detail I had when I was researching the same topic for GB3:

[Part 1](https://medium.com/@drevets/you-cant-prompt-a-file-download-with-the-content-disposition-header-using-axios-xhr-sorry-56577aa706d6)
[Part 2](https://medium.com/@nerdyman/prompt-a-file-download-with-the-content-disposition-header-using-fetch-and-filesaver-8683faf565d0)


Other than that, it works fine

TIL-EBP (changes requested):

libs/asset-viewer/src/lib/components/asset-search-detail/asset-search-detail.component.ts
daniel-va: Filesaver does internally the same as I did here with plain JS. It offers - to my knowledge - no security advantages over just downloading the file ourselves.

The topic that both articles talk about is just the `Content-Disposition` header, which tells the browser how to handle a downloaded file. It is not directly related to security, but rather how we would want to user to see and receive a file.

Note that Filesaver does not handle the `Content-Disposition` header either, we would have to do this ourselves, just like we would have to do here. Since we can cover all of our use cases without handling that header, I would argue we can just leave the solution as-is.

TIL-EBP (approved):

PR #181, opened 2024-06-24
feature 147: Added edit-asset test

DmByK:
Because the delete function is not ready yet, this pr contains only the edit asset test.

### DIFF
--- a/e2e/cypress/e2e/edit-asset.feature
+++ b/e2e/cypress/e2e/edit-asset.feature
@@ -1,0 +1,17 @@
+Feature: edit
+
+  Background:
+    Given The user is logged in
+    And Test asset is created
+    And Elastic index is created
+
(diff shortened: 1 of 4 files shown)


e2e/cypress/e2e/edit-asset.feature
```diff
@@ -0,0 +1,17 @@
...
+
+  Background:
+    Given The user is logged in
+    And Test asset is created
+    And Elastic index is created
+
+  Scenario: Edit asset
+    When A user types CypressTestAsset into the search field
+    And The user clicks enter
+    And The user should see the asset in results
```
daniel-va, line 11: Shouldn't this be a `Then`?
```suggestion
    Then The user should see the asset in results
```

daniel-va (approved):

PR #195, opened 2024-07-03
Feature 27: Selection of Points

### LINKED ISSUE #27: Selection of points without having to draw polygons
Currently, point features cannot be clicked on the map. They must be captured using the "Polygon selection" tool. 
As a user, I'd like to directly select an asset on the map without using the "polygon selection"

Question Rafael: Point selection as tool in toolbox (next to polygon selection)? Or is the point selection always active?

Direct or point selection tool next to polygon selection tool in toolbox would be nice. So the selection method can be chosen by the user itself. Before selecting one of the tools needs to be chosen. Otherwise there might be a conflict with the pan-function in the map (erroneous click on point feature instead of panning by user).

- Point selection
- "rectangle" mode (wie map.geo.admin); Ctrl. + Mouse

See also Issue #24 (backlog)!

**Acceptance criteria**
1. When clicking on an object in the map, the "Detailinformation"-Panel opens for the selected object.
2. In the toolbar there is a rectangle selection tool. It is also usable via Ctrl + Mouse.
3. When points are selected (e.g. via polygon tool), blue points are still selectable via e.g. clicking.

daniel-va:
Depends on #6.
Resolves #27 
Allows study points on the map to be clicked in order to select the asset that they belong to.

### DIFF
--- a/apps/server-asset-sg/src/features/asset-old/asset.service.ts
+++ b/apps/server-asset-sg/src/features/asset-old/asset.service.ts
@@ -44,13 +44,24 @@ export class AssetService {
   }
 
   async getAllStudies() {
-    const rawData: { studyId: string; isPoint: boolean; centroidGeomText: string }[] =
-      await this.prismaService.$queryRawUnsafe(
-        'select study_id as "studyId", is_point as "isPoint", centroid_geom_text as "centroidGeomText" from public.all_study'
-      );
+    interface RawStudy {
+      studyId: string;
+      assetId: number;
+      isPoint: boolean;
+      centroidGeomText: string;
+    }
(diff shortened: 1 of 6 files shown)


TIL-EBP:
These are just some things i found when testing the aplication, before looking at any of the code (also, I am only revieweing this PR, not the one this depends on):

- Remove the openLAyer map controls
- The map does no longer zoom to the search results upon completing a search
- selecting assets on the map from the search results no longer works
- not part of your PR, but when resetting a polygon from the search (without a search term), an empty saerch gets triggered

TIL-EBP (changes requested):

TIL-EBP (approved):

PR #227, opened 2024-07-18
Feature/asset 160 workgroups admin crud

### LINKED ISSUE #160: Workgroups Admin CRUD
**Beschreibung**

Über eine Admin Seite kann ein Admin neue workgroups erstellen, bearbeiten und deaktivieren.
Es gibt ausserdem eine Auflistung aller Benutzerinnen und deren Berechtigungen.
Die Berechtigungen einer Benutzerin können bearbeitet werden.

Nice to have: Filter nach Workgroup

**Acceptance Criteria**

 - [ ] Workgroups und Benutzer müssen verwaltet werden können.

**UI Vorschlag**
Den folgenden Vorschlag mit dem UX Team abgleichen, sobald sie verfügbar sind.

Die Wireframes beschreiben ausschliesslich Funktionalität - es steht dem Entwickler frei, eine geeignete Komponente zu wählen.
Image
Image

TIL-EBP:
resolves  #160 

Open issues: 
Design not final
Scrolling of tables


libs/admin/src/lib/components/users/users.component.html
```diff
@@ -1,39 +1,49 @@
...
       <td mat-cell *matCellDef="let user">{{ user.lang }}</td>
     </ng-container>
+    <ng-container matColumnDef="workgroups">
+      <th mat-header-cell *matHeaderCellDef translate>admin.workgroups</th>
+      <td
+        mat-cell
+        *matCellDef="let user"
+        [matTooltip]="formatWorkgroupsTooltip(user.workgroups)"
+        matTooltipClass="workgroups-tooltip"
+      >
```
daniel-va, line 28: Position the tooltip left so it appears right next to the text.
```suggestion
      <td
        mat-cell
        *matCellDef="let user"
        [matTooltip]="formatWorkgroupsTooltip(user.workgroups)"
        matTooltipPosition="left"
        matTooltipClass="workgroups-tooltip"
      >
```

libs/admin/src/lib/components/workgroup-edit/workgroup-edit-component.ts
```diff
@@ -1,62 +1,177 @@
...
-export class WorkgroupEditComponent implements OnInit {
-  private _adminService = inject(AdminService);
-
-  @Input() workgroup: Workgroup | null = null;
-  @Input() users: User[] = [];
-  @Input() mode: 'edit' | 'create' | undefined = undefined;
+export class WorkgroupEditComponent implements OnInit, OnDestroy {
+  public workgroup: Workgroup | undefined;
+  public mode: Mode = 'edit';
+  public roles: Role[] = Object.values(Role);
```
daniel-va, line 27:
```suggestion
  public readonly roles: Role[] = Object.values(Role);
```

libs/admin/src/lib/components/workgroup-edit/workgroup-edit-component.ts
```diff
@@ -1,62 +1,177 @@
...
+              id: 0,
+              name: '',
+              users: [],
+              assets: [],
+              disabled_at: null,
+            };
+          }
+        })
+      )
+      .subscribe();
```
daniel-va, line 63: Can we move this into its own method?

libs/admin/src/lib/components/workgroup-edit/workgroup-edit-component.ts
```diff
@@ -1,62 +1,177 @@
...
+
+    this.initSubscriptions();
+  }
+
+  public ngOnDestroy() {
+    this.store.dispatch(actions.resetWorkgroup());
```
daniel-va, line 69: Missing unsubscribe.

libs/admin/src/lib/components/workgroup-edit/workgroup-edit-component.ts
```diff
@@ -1,62 +1,177 @@
...
+        assets: this.workgroup.assets,
+        disabled_at: this.workgroup.disabled_at,
+        users: this.workgroup.users.filter((u) => u.userId !== user.userId),
+      };
+      this.updateWorkgroup(this.workgroup.id, updatedWorkgroup);
+    }
+  }
+
+  public cancel() {
+    this.router.navigate(['../'], { relativeTo: this.route });
```
daniel-va, line 127: Prefix this call with `void`, as it returns a promise which we want to ignore.
```suggestion
    void this.router.navigate(['../'], { relativeTo: this.route });
```

libs/admin/src/lib/components/workgroup-edit/workgroup-edit-component.ts
```diff
@@ -1,62 +1,177 @@
...
+    this.store.dispatch(actions.resetWorkgroup());
+  }
+
+  public createWorkGroup() {
+    if (this.workgroup && this.formGroup.valid) {
```
daniel-va, line 132: Invert this `if`.

libs/admin/src/lib/components/workgroup-edit/workgroup-edit-component.ts
```diff
@@ -1,62 +1,177 @@
...
+      this.workgroup$
+        .pipe(
+          tap((workgroup) => {
+            if (workgroup) {
+              this.workgroup = workgroup;
+              this.initializeForm(workgroup);
+            }
+          })
+        )
+        .subscribe()
```
daniel-va, line 160: Merge `tap` into `subscribe`.

libs/admin/src/lib/components/workgroup-edit/workgroup-edit-component.ts
```diff
@@ -1,62 +1,177 @@
...
+    );
+
+    this.subscriptions.add(
+      this.formGroup.valueChanges.pipe(debounceTime(300)).subscribe((value) => {
+        if (this.workgroup) {
```
daniel-va, line 165: Invert this `if`.

libs/admin/src/lib/components/workgroup-edit/workgroup-edit.component.scss
```diff
@@ -1,4 +1,37 @@
-.form {
-  display: flex;
-  flex-direction: column;
+@use "../../styles/variables";
+
+.workgroup-edit {
```
daniel-va, line 3: Just like before, remove this class.

libs/admin/src/lib/components/workgroups/workgroups.component.scss
```diff
@@ -0,0 +1,30 @@
+@use "../../styles/variables";
+
+.workgroups {
```
daniel-va, line 3: Just like before, remove this class.

libs/admin/src/lib/services/admin.service.ts
```diff
@@ -55,10 +65,18 @@ export class AdminService {
     return this._httpClient.get('/api/admin/user').pipe(map((res) => res as User[]));
   }
 
+  public getUserNew(id: string): Observable<User> {
```
daniel-va, line 68: Why is there a `new` prefix on these methods? Can't we just replace the old methods now?

libs/admin/src/lib/services/admin.service.ts
```diff
@@ -55,10 +65,18 @@ export class AdminService {
...
+    return this._httpClient.get(`/api/users/${id}`).pipe(map((res) => res as User));
+  }
+
   public getWorkgroups(): Observable<Workgroup[]> {
     return this._httpClient.get('/api/workgroups').pipe(map((res) => res as Workgroup[]));
   }
 
+  public getWorkgroup(id: string): Observable<Workgroup> {
+    return this._httpClient.get(`/api/workgroups/${id}`).pipe(map((res) => res as Workgroup));
```
daniel-va, line 77: Mapping responses just to cast them is not necessary - we can just specify the type in the request:
```suggestion
    return this._httpClient.get<Workgroup>(`/api/workgroups/${id}`);
```
Note that type casts are a no-op when TS is compiled to JS - `a as B` is compiled to just `a`.

libs/admin/src/lib/state/admin.actions.ts
```diff
@@ -0,0 +1,74 @@
+import { createAction, props } from '@ngrx/store';
+import { User, Workgroup } from '../services/admin.service';
+
+export const findUser = createAction(
+  '[Asset Search] Find user',
```
daniel-va, line 5: Replace the prefix on all of these actions - we are not in the asset search state here.
```suggestion
  '[Admin] Find user',
```

libs/admin/src/lib/state/admin.actions.ts
```diff
@@ -0,0 +1,74 @@
...
+  props<{
+    workgroups: Workgroup[];
+  }>()
+);
+
+export const updateWorkgroup = createAction(
+  '[Asset Search] Update Workgroup',
+  props<{
+    workgroupId: number;
+    workgroup: Omit<Workgroup, 'id'>;
```
daniel-va, line 61: Why do we omit the `id`, but then specify it as an additional field?

libs/admin/src/lib/state/admin.effects.ts
```diff
@@ -0,0 +1,118 @@
...
+import * as D from 'io-ts/Decoder';
+import { filter, map, merge, switchMap, tap } from 'rxjs';
+
+import { AdminService, User, Workgroup } from '../services/admin.service';
+import * as actions from './admin.actions';
+import { AdminState, AppStateWithAdmin } from './admin.reducer';
+
+@UntilDestroy()
+@Injectable()
+export class AssetAdminEffects {
```
daniel-va, line 18: We can just call this `AdminEffects`, I don't think the prefix adds anything of value.
```suggestion
export class AdminEffects {
```

libs/admin/src/lib/state/admin.effects.ts
```diff
@@ -0,0 +1,118 @@
...
+  private readonly store = inject(Store<AdminState>);
+  private readonly actions$ = inject(Actions);
+  private readonly adminService = inject(AdminService);
+  private readonly adminStore = inject(Store<AppStateWithAdmin>);
```
daniel-va, line 22: This variable is unused.

libs/admin/src/lib/state/admin.effects.ts
```diff
@@ -0,0 +1,118 @@
...
+  constructor() {
+    merge(this.store.select(fromAppShared.selectRDReferenceData))
+      .pipe(
+        filter(RD.isFailure),
+        map((e) => e.error),
+        filter(isDecodeError)
+      )
+      .subscribe((e) => {
+        console.error('DecodeError', D.draw(e.cause));
+      });
```
daniel-va, line 35: What is this for?

libs/admin/src/lib/state/admin.effects.ts
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  public findUser$ = createEffect(() => {
+    return this.actions$.pipe(
+      ofType(actions.findUser),
+      switchMap(({ userId }) => {
+        return this.adminService.getUserNew(userId).pipe(map((user: any) => actions.setUser({ user })));
```
daniel-va, line 42: Don't use `any` - its basically never a good idea, and can actually just be removed here.
```suggestion
        return this.adminService.getUserNew(userId).pipe(map((user) => actions.setUser({ user })));
```

libs/admin/src/lib/state/admin.effects.ts
```diff
@@ -0,0 +1,118 @@
...
+        map((e) => e.error),
+        filter(isDecodeError)
+      )
+      .subscribe((e) => {
+        console.error('DecodeError', D.draw(e.cause));
+      });
+  }
+
+  public findUser$ = createEffect(() => {
+    return this.actions$.pipe(
```
daniel-va, line 39: We don't need a method body and `return` statement for all effects in this file.
```suggestion
  public findUser$ = createEffect(() => this.actions$.pipe(
```

libs/admin/src/lib/state/admin.effects.ts
```diff
@@ -0,0 +1,118 @@
...
+      })
+    );
+  });
+
+  public createWorkgroup$ = createEffect(
+    () => {
+      return this.actions$.pipe(
+        ofType(actions.createWorkgroup),
+        switchMap(({ workgroup }) => this.adminService.createWorkgroup(workgroup)),
+        tap((workgroup) => this.router.navigate([`/de/admin/workgroups/${workgroup.id}`], { relativeTo: this.route }))
```
daniel-va, line 78: Use `switchMap` instead of `tap` so the navigation's promise gets resolved by NgRX.
```suggestion
        switchMap((workgroup) => this.router.navigate([`/de/admin/workgroups/${workgroup.id}`], { relativeTo: this.route }))
```

libs/admin/src/lib/state/admin.reducer.ts
```diff
@@ -0,0 +1,100 @@
...
+  Loading = 'loading',
+  Loaded = 'loaded',
+}
+
+export interface AdminState {
+  selectedWorkgroup: Workgroup | undefined;
+  workgroups: Workgroup[];
+  selectedUser: User | undefined;
+  users: User[];
+  loadingState: LoadingState;
```
daniel-va, line 16: Can't we just make this a `isLoading: boolean`?

libs/admin/src/lib/state/admin.reducer.ts
```diff
@@ -0,0 +1,100 @@
...
+export interface AppStateWithAdmin extends AppState {
+  admin: AdminState;
+}
+
+const initialState: AdminState = {
+  selectedWorkgroup: undefined,
+  workgroups: [],
+  selectedUser: undefined,
+  users: [],
+  loadingState: LoadingState.Loaded,
```
daniel-va, line 28: Why do we start with `Loaded` instead of `Loading`?

libs/admin/src/lib/state/admin.selector.ts
```diff
@@ -0,0 +1,11 @@
+import { createSelector } from '@ngrx/store';
+import { AppStateWithAdmin } from './admin.reducer';
+
+const assetAdminFeature = (state: AppStateWithAdmin) => state.admin;
```
daniel-va, line 4: I'd drop the `asset` prefix here, the store also doesn't use it.
```suggestion
const adminFeature = (state: AppStateWithAdmin) => state.admin;
```

daniel-va (changes requested):

apps/server-asset-sg/src/app.controller.ts
```diff
@@ -89,6 +88,15 @@ export class AppController {
     // deprecated
   }
 
+  /**
+   * @deprecated
+   */
+  @Get('/admin/user/:id')
+  @Redirect('../users/:id', 301)
+  getUserById() {
```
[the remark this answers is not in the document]
TIL-EBP: At firsrt i wanted to use all the same endpoints until we switch to the new endpoint for all of them, but I removed it again

libs/admin/src/lib/components/add-users/add-users.component.ts
```diff
@@ -0,0 +1,106 @@
...
+  public formGroup: FormGroup = new FormGroup({
+    users: new FormControl([], Validators.required),
+    role: new FormControl(Role.Viewer, Validators.required),
+  });
+
+  public roles: Role[] = Object.values(Role);
+  public users: User[] = [];
+  public usersToAdd: UserOnWorkgroup[] = [];
+  public workgroup: Workgroup | undefined;
+  public mode: Mode = 'edit';
```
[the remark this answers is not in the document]
TIL-EBP: All fileds except the roles are updates, so its to only one that can be made readonly, right?

libs/admin/src/lib/components/add-users/add-users.component.ts
vej-ananas, line 1: Selecting the role resets the user selection and even reenables disabled ones.

libs/admin/src/lib/components/user-edit/user-edit.component.html
```diff
@@ -1,44 +1,103 @@
...
-        <mat-option value="en">EN</mat-option>
-        <mat-option value="fr">FR</mat-option>
-        <mat-option value="it">IT</mat-option>
-        <mat-option value="rm">RM</mat-option>
-      </mat-select>
-    </mat-form-field>
+    <div class="user-edit__workgroups">
+      <div class="user-edit__workgroups__header">
+        <h2 translate>admin.workgroups</h2>
+        <form>
```
vej-ananas, line 29: Already included workgroups can be selected again. They should be disabled instead.
`admin.name` is visible in the component.
After workgroup selection, the input should be empty again.

libs/admin/src/lib/components/workgroups/workgroups.component.html
```diff
@@ -1,31 +1,41 @@
...
     </ng-container>
+
     <ng-container matColumnDef="actions">
-      <th mat-header-cell *matHeaderCellDef>Aktionen</th>
+      <th mat-header-cell *matHeaderCellDef translate>admin.actions</th>
       <td mat-cell *matCellDef="let workgroup">
-        <button mat-button (click)="selectWorkgroup(workgroup)">Editieren</button>
-        <button mat-flat-button>Benutzer hinzufügen</button>
-        <button mat-icon-button (click)="deactivateWorkgroup(workgroup)">Deaktivieren</button>
+        <div class="workgroups__table__actions">
```
vej-ananas, line 27: Do not offer more than one action in the column but a `...` menu instead.
Also make sure that the last few columns dont scale in size but remain small - the `User` column is the one to take the space

libs/admin/src/lib/components/user-edit/user-edit.component.html
```diff
@@ -1,44 +1,103 @@
...
+            <td mat-cell *matCellDef="let workgroup">{{ workgroup.workgroup.name }}</td>
+          </ng-container>
 
-    <mat-checkbox formControlName="isAdmin" color="primary">Admin</mat-checkbox>
-    <div>Workgroups</div>
+          <!-- Role Column -->
+          <ng-container matColumnDef="role">
+            <th mat-header-cell *matHeaderCellDef translate>admin.role</th>
+            <td mat-cell *matCellDef="let workgroup">
+              <mat-form-field appearance="fill">
```
vej-ananas, line 73: Adding `subscriptSizing='dynamic'` makes the ui look nicer

libs/admin/src/lib/state/admin.actions.ts
TIL-EBP: The WorkgroupDataBoundary does not allow an id field, so we just send the rest of the body. And instead of mapping the workgroup without the id in the effect or the service, it is done in the component. But i can also remove the id inside the service method

libs/admin/src/lib/components/users/users.component.scss
```diff
@@ -0,0 +1,26 @@
...
+    background-color: variables.$grey-03;
+  }
+
+  .mat-mdc-header-row .mat-mdc-cell {
+    background-color: variables.$grey-01;
+  }
+
+  .mat-mdc-row:hover .mat-mdc-cell {
+    background-color: variables.$grey-01;
+  }
```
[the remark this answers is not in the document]
TIL-EBP: You are correct about the top one, but the hover does work

libs/admin/src/lib/components/add-users/add-users.component.ts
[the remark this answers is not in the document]
daniel-va: yes.

libs/admin/src/lib/state/admin.actions.ts
daniel-va: So the `workgroup` field actually always has the type `WorkgroupData`? Because if so, then we can just use that type instead of `Omit<...>`.

libs/admin/src/lib/components/users/users.component.scss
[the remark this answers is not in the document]
daniel-va: I don't see how, but if it works, it works I guess.

libs/admin/src/lib/components/user-edit/user-edit.component.html
```diff
@@ -1,44 +1,103 @@
-<form [formGroup]="formGroup">
-  <div class="edit-area">
+<div class="user-edit">
+  <form [formGroup]="formGroup">
     <div *ngIf="user" class="header">
-      <div class="email">
+      <h2 class="email">
         {{ user.email }}
```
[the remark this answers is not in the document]
TIL-EBP: I don't think there is, in the mockup, this is not an editable field

libs/admin/src/lib/components/add-users/add-users.component.ts
```diff
@@ -0,0 +1,106 @@
...
+      }
+      this.dialogRef.close();
+    }
+  }
+
+  public isUserInWorkgroupWithSelectedRole(userId: string): boolean {
+    return (
+      this.workgroup?.users.some(
+        (user) => user.userId === userId && user.role === this.formGroup.controls['role'].value
+      ) ?? false
```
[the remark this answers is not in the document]
TIL-EBP: After refactoring, I no l onger have to check for the role. it is now simply 
`public isUserInWorkgroup(userId: string): boolean {
    return this.workgroup?.users.some((user) => user.userId === userId) ?? false;
  }`

apps/server-asset-sg/src/features/users/user.repo.ts
```diff
@@ -66,7 +66,7 @@ export class UserRepo implements Repo<User, UserId, UserData & { oidcId: string
           isAdmin: data.isAdmin,
           workgroups: {
             deleteMany: {
-              workgroupId: { notIn: data.workgroups?.map((workgroup) => workgroup.workgroupId) },
```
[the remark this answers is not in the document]
TIL-EBP: I got a little confused by this as well, but the reason i changed this was because the original code would not allow us to update the role of an exisiting workgroup, as it did not get deleted and then skipped in the next step. 
So if we delete all the workgroups related to a user and then add them all again, we can update as well as add new ones

apps/server-asset-sg/src/features/workgroups/workgroup.repo.ts
```diff
@@ -64,7 +64,7 @@ export class WorkgroupRepo implements Repo<Workgroup, WorkgroupId, WorkgroupData
           users: data.users
             ? {
                 deleteMany: {
-                  userId: { notIn: data.users.map((user) => user.userId) },
```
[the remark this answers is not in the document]
TIL-EBP: Same as above, we need to be able to update the roles of a user, and with the original code, this does not work

libs/admin/src/lib/state/admin.actions.ts
TIL-EBP: Agreed. This to me again raises the question where to define all of these types. Currenlty, we have them in the model in the backend and the service in the frontend, which is suboptimal. Where should we move them so we dont dublicate?

daniel-va: In #159, I already moved all the types into the new `libs/shared2` library. Since we need to merge the two branches anyways later on this week, you could maybe add a TODO that the type needs to be changed to `WorkgroupData`?

libs/admin/src/lib/components/user-edit/user-edit.component.ts
```diff
@@ -1,43 +1,164 @@
...
+      this.updateUser({ ...this.user, workgroups: updatedWorkgroups });
+    }
+  }
+
+  public deleteWorkgroupFromUser(workgroup: WorkgroupOnUser) {
+    if (this.user) {
+      const updatedWorkgroups = this.user.workgroups.filter(
+        (userWorkgroup) => userWorkgroup.workgroupId !== workgroup.workgroupId
+      );
+      this.updateUser({ ...this.user, workgroups: updatedWorkgroups });
```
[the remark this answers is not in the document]
TIL-EBP: Will be done after merging back to _workgroups_ branch

libs/admin/src/lib/state/admin.actions.ts
TIL-EBP: Will move the types to shared2 after everything is merged to avoid conflicts

apps/server-asset-sg/src/features/users/user.repo.ts
```diff
@@ -66,7 +66,7 @@ export class UserRepo implements Repo<User, UserId, UserData & { oidcId: string
...
+              workgroupId: {},
```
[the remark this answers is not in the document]
TIL-EBP: We ned to delete all UserWorkgroups for this user, not just the ones that have been removed, because their roles can also be updated, and this changes was not accounted for with the previous code

libs/admin/src/lib/components/add-users/add-workgroup-user-dialog.component.scss
```diff
@@ -0,0 +1,13 @@
...
+  padding: 20px 24px 0 24px;
+}
+
+.select {
+  width: 100%;
+}
+
+.actions {
+  display: flex;
+  justify-content: end;
```
daniel-va, line 11: Prefer `flex-start` and `flex-end` over `start` and `end` in flex layouts, as the latter two are not as widely supported.
```suggestion
  justify-content: flex-end;
```

libs/admin/src/lib/components/users/users.component.scss
```diff
@@ -0,0 +1,32 @@
...
+.users {
+  height: 100%;
+  display: flex;
+  flex-direction: column;
+  padding: 1rem;
+}
+
+.users__header {
+  display: flex;
+  justify-content: start;
```
daniel-va, line 16: Prefer `flex-start` and `flex-end` over `start` and `end` in flex layouts, as the latter two are not as widely supported.
```suggestion
  justify-content: flex-start;
```

libs/admin/src/lib/components/users/users.component.scss
```diff
@@ -0,0 +1,32 @@
...
+  align-items: center;
+  margin-bottom: 1rem;
+}
+
+.users__table {
```
daniel-va, line 21: Remove the `users__` prefix on all classes in this component as every element in this component's template is wrapped inside the `.users` div anyways.

libs/admin/src/lib/components/workgroup-edit/workgroup-edit.component.scss
```diff
@@ -0,0 +1,55 @@
...
+  padding: 1.5rem;
+
+  .workgroup-edit__form {
```
daniel-va, line 6: Remove the `.workgroup-edit__` prefix on all classes in this component as every element in this component's template is wrapped inside the `.workgroup-edit` div anyways.

For the same reasons, you can also move all classes nested within `.workgroup-edit` to the CSS's top level, as nesting them does not change how they are applied anyways.

libs/admin/src/lib/components/workgroups/workgroups.component.scss
```diff
@@ -0,0 +1,31 @@
...
+  height: 100%;
+  display: flex;
+  flex-direction: column;
+
+  .workgroups__header {
```
daniel-va, line 8: Remove the `workgroups__` prefix on all classes in this component as every element in this component's template is wrapped inside the `.workgroups` div anyways.

daniel-va:
Merged in #239.

PR #237, opened 2024-07-23
Fix: Add missing optional NX dependencies used by docker build

daniel-va:
The `develop` build is missing some NX build dependencies used when building the app's Docker images. This PR adds them.


vej-ananas (approved):
Kind of embarrassing that we have to add these manually..

daniel-va:
> Kind of embarrassing that we have to add these manually..

@vej-ananas Seems to be a bug with NX when used with Node v22. It might already be fixed in the newest version of NX (our version is `19.2.1`, stable is `19.5.2`).

PR #261, opened 2024-08-30
Automatically sync search index

vej-ananas:
#247

### DIFF
--- a/apps/server-asset-sg/src/features/assets/sync/asset-sync.service.ts
+++ b/apps/server-asset-sg/src/features/assets/sync/asset-sync.service.ts
@@ -1,0 +1,90 @@
+import fs from 'fs/promises';
+import { Injectable, OnApplicationBootstrap } from '@nestjs/common';
+import { SchedulerRegistry } from '@nestjs/schedule';
+import { CronJob } from 'cron';
+import { AssetRepo } from '@/features/assets/asset.repo';
+import { AssetSearchService } from '@/features/assets/search/asset-search.service';
+
+@Injectable()
+export class AssetSyncService implements OnApplicationBootstrap {
+  constructor(
+    private readonly assetSearchService: AssetSearchService,
+    private readonly schedulerRegistry: SchedulerRegistry,
+    private readonly assetRepo: AssetRepo
+  ) {}
+  async onApplicationBootstrap() {
+    const syncFileExists = await this.isSyncRunning();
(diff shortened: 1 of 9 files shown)


apps/server-asset-sg/src/features/assets/sync/asset-sync.service.ts
```diff
@@ -0,0 +1,90 @@
...
+    setTimeout(async () => {
+      await this.assetSearchService.syncWithDatabase(writeProgress);
+      await fs.rm(assetSyncFile);
+    });
+  }
+
+  private async startSyncIfIndexOutOfSync() {
+    console.debug(`startSyncIfIndexOutOfSync.`);
+    const numberOfAssets = await this.assetRepo.count();
+    const numberOfIndixedAssets = await this.assetSearchService.count();
```
TIL-EBP, line 73:
```suggestion
    const numberOfIndexedAssets = await this.assetSearchService.count();
```

apps/server-asset-sg/src/features/assets/sync/asset-sync.service.ts
```diff
@@ -0,0 +1,90 @@
...
+    const syncFileExists = await this.isSyncRunning();
+    if (syncFileExists) {
+      void fs.rm(assetSyncFile);
+    }
+
+    if (process.env.ANONYMOUS_MODE === 'true') {
+      console.log('Anonymous Mode is activated. Search Index will be automatically synced.');
+      await this.startSyncIfIndexOutOfSync();
+
+      const every20Minutes = '*/20 * * * *';
```
TIL-EBP, line 25: Isn't this a bit much? As far as I understood, the data is only updated once a day dunring the night, so there shouldn't be a need to update the sync every 20 minutes, right?

package.json
```diff
@@ -61,6 +61,7 @@
     "cache-manager": "^5.4.0",
     "class-transformer": "^0.5.1",
     "class-validator": "^0.14.1",
+    "cron": "^3.1.7",
```
TIL-EBP, line 64: There is a '@nestjs/schedule' [module ](https://docs.nestjs.com/techniques/task-scheduling)which does the same thing (probably uses cron under the hood), so we could consider using this instead of cron directly, just to make the most of using nestjs

TIL-EBP (approved):

PR #267, opened 2024-09-05
Refactor Search flow in NGRX to fix routing

### LINKED ISSUE #240: Bug: Suche und URL out of Sync
**Beschreibung**
Die Suchfilter und URL gehen out of Sync, wenn man auf andere Seiten navigiert.

Wenn man eine , auf eine andere Seite wie Asset Management navigiert und wieder auf Asset Map klickt, dann 

1. Suche starten z.B. durch einen Suchfilter
2. Auf eine andere Seite wie Asset Management navigieren
3. Über Sidebar wieder auf Asset Map navigieren
4. Filter sind gesetzt wie zuvor, aber die URL ist leer `https://dev-assets.swissgeol.ch/de`

**Beobachtetes Verhalten**
URL leert sich durch navigation

**Erwartetes Verhalten**
URL und Filter stellen ein einheitliches Bild dar.

vej-ananas:
Resolves #240


libs/asset-viewer/src/lib/state/asset-search/asset-search.effects.ts
```diff
@@ -164,53 +88,106 @@ export class AssetSearchEffects {
...
+    this.queryParams$.pipe(
+      filter(({ query }) => Object.values(query).some((value) => value !== undefined)),
+      map(({ query }) => actions.search({ query: query }))
+    )
+  );
+
+  public readAssetIdQueryParam$ = createEffect(() =>
+    this.queryParams$.pipe(
+      map(({ assetId }) => assetId),
+      filterNullish(),
```
TIL-EBP, line 155: is this cutom method really needed? If so, are there other cases wehre it could be used?

TIL-EBP (approved):
Some general remark: 

When  I start a search from the map page, then navigate to "Verrwaltung" and back to the map, the filters are closed. When I do the same via Benutzerverwaltung, the filters are open (as they should be IMO) This probably still has something to do with the portals which are still used in the "Verwatltungs"-Tab?? 

Nitpick: When hiding the results table, navigating away and back to map, the table is now open again. previously, this state was persisted

When I search by polygon and then reset the polygon ("Polygon aufheben") it takes much longer fot the search result to disappear than before. When going vis "Suche zurücksetzen", it is (still) almost instant

Not directly part of your PR, but i notived that when selecting an asset, navigatng to a different page and then navigtaing back to the map page, the asset is still selected, but the map zoom is not on focussed on this asset but the entire map. What is the desired behaviour there?

libs/client-shared/src/lib/i18n.ts
```diff
@@ -3,6 +3,10 @@ import { Observable, of } from 'rxjs';
 
 export type Langs = 'de' | 'en' | 'fr' | 'it' | 'rm';
 
+export function isSupportedLang(lang: string): lang is Langs {
+  return ['de', 'en', 'fr', 'it', 'rm'].includes(lang);
```
TIL-EBP, line 7: Is there a simple way of doing this without having to dublicate the array?

ChatGPT suggests this:
```

export const supportedLangs = ['de', 'en', 'fr', 'it', 'rm'] as const;

export type Langs = typeof supportedLangs[number];

export function isSupportedLang(lang: string): lang is Langs {
  return supportedLangs.includes(lang as Langs);
}

```

libs/asset-viewer/src/lib/state/asset-search/asset-search.actions.ts
```diff
@@ -35,4 +36,4 @@ export const closeFilters = createAction('[Asset Viewer] Close Filters');
 export const openResults = createAction('[Asset Viewer] Open Results');
 export const closeResults = createAction('[Asset Viewer] Close Results');
 export const setStudies = createAction('[Asset Viewer] Load Studies', props<{ studies: AllStudyDTOs }>());
-export const loadSearch = createAction('[Asset Viewer] Load Search', props<{ query: AssetSearchQuery }>());
+export const loadSearch = createAction('[Asset Search] Load Search', props<{ query: AssetSearchQuery }>());
```
TIL-EBP, line 39: I believe the convetion is to have the top level name be the same for all actions. So theoretically, all actions should be called one or the other (Assert Search or Asset Viewer) or we can split them into separate files

libs/asset-viewer/src/lib/state/asset-search/asset-search.effects.ts
```diff
@@ -141,27 +154,37 @@ export class AssetSearchEffects {
...
   );
 
-  // noinspection JSUnusedGlobalSymbols
   public readSearchQueryParams$ = createEffect(() =>
     this.queryParams$.pipe(
-      filter(({ query }) => Object.values(query).some((value) => value !== undefined)),
-      map(({ query }) => actions.search({ query: query }))
+      withLatestFrom(this.store.select(selectAssetSearchQuery)),
+      filter(([params, query]) => !deepEqual(params.query, query)),
+      map(([params, storeQuery]) => {
```
TIL-EBP, line 161: same here as below, i would prefer the same variables to have the same name in the entire pipe

libs/asset-viewer/src/lib/components/map/map.component.ts
```diff
@@ -198,6 +204,10 @@ export class MapComponent implements AfterViewInit, OnChanges, OnDestroy {
     }
   }
 
+  private equalsCheck<T>(a: T[] | null, b: T[] | null) {
```
TIL-EBP, line 207: Since this is generic it coiuld live inside its own file as it does not really have anything to do withthe map per se

apps/client-asset-sg/src/app/components/menu-bar/menu-bar.component.html
TIL-EBP, line 1: maybe also change the last button inside to an anchor so it is all consistent

TIL-EBP: and the aria-label

libs/asset-viewer/src/lib/state/asset-search/asset-search.effects.ts
```diff
@@ -141,27 +154,37 @@ export class AssetSearchEffects {
...
   );
 
   public readAssetIdQueryParam$ = createEffect(() =>
     this.queryParams$.pipe(
-      map(({ assetId }) => assetId),
-      filterNullish(),
-      map((assetId) => actions.assetClicked({ assetId }))
+      withLatestFrom(this.store.select(selectCurrentAssetDetail)),
+      filter(([params, assetDetail]) => params.assetId !== assetDetail?.assetId),
+      map(([params, storeAsset]) => {
```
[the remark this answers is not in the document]
vej-ananas: We check the presence of any parameters at all because we want to decide whether to use the data from the store or from the query parameters

TIL-EBP (approved):

PR #271, opened 2024-09-16
Create data pipelines for PROD View

### LINKED ISSUE #269: Data Pipelines for PROD View and Extern
**Beschreibung**
Für die Synchronisierung von Daten zwischen PROD und PROD View / Extern müssen Pipelines eingerichtet werden.
Dafür wird eine neue Applikation mit eigenem Docker Image geschrieben, die ins Kubernetes Cluster deployed und  entsprechend konfiguriert werden kann.

vej-ananas:
resolves #269


apps/pipeline-asset-sg/src/export-to-view.service.ts
```diff
@@ -0,0 +1,334 @@
...
+    log(`Starting file export.`);
+    // Read all unique file ids from the assetFiles
+    const assetFiles = await this.sourcePrisma.assetFile.findMany({ where: { assetId: { in: assetIds } } });
+    const fileIds = assetFiles.map((af) => af.fileId).filter(unique);
+    const files = await this.sourcePrisma.file.findMany({ where: { fileId: { in: fileIds } } });
+
+    // Filter out files which start with 'LDoc'
+    const filesWithoutLegalDocs = files.filter((file) => !file.fileName.includes('LDoc'));
+    const fileIdsWithoutLegalDocs = filesWithoutLegalDocs.map((file) => file.fileId);
+    const assetFilesWithoutLegalDocs = assetFiles.filter((af) => fileIdsWithoutLegalDocs.includes(af.fileId));
```
daniel-va, line 248: As I see it, the `"LDoc"` filter could just be moved into the initial `findMany`:
```ts
const assetFiles = await this.sourcePrisma.assetFile.findMany({
  where: {
    assetId: { in: assetIds },
    file: { fileName: { not: { contains: 'LDoc' } } },
  },
});
```

apps/pipeline-asset-sg/src/export-to-view.service.ts
```diff
@@ -0,0 +1,334 @@
...
+    const result = await this.destinationPrisma.$executeRaw`
+      INSERT INTO study_location (study_location_id, asset_id, geom_quality_item_code, geom)
+      VALUES ${Prisma.join(formattedStudyLocations)}
+    `;
+
+    log(`Created ${result} study locations.`);
+  }
+
+  /**
+   * Export study traces as unsafe query with st_astext.
```
daniel-va, line 118: Just a quick note that Prisma also has _unsafe_ raw queries, so this comment is a bit misleading, as in fact, we're using _safe_ raw queries. The safety part is just refering to whether the query is a simple, unchecked string (unsafe) or an instance of `Prisma.Sql` (safe).

apps/pipeline-asset-sg/src/log.ts
daniel-va, line 1: I would move the contents of this file into `utils.ts`, since both files seem to be very small and have a similiar purpose.

apps/pipeline-asset-sg/src/reset-and-migrate-database.ts
daniel-va, line 1: Similar to `log`, I would merge this file into `utils.ts`. I don't really see the purpose of so many files with only a single function in them.

apps/server-asset-sg/docker/Dockerfile
```diff
@@ -8,7 +8,7 @@ ENV NODE_ENV=development
 WORKDIR /app
 COPY . .
 
-RUN npm install && ./node_modules/.bin/prisma generate
```
daniel-va, line 11: Why is this no longer needed?

daniel-va (changes requested):

apps/pipeline-asset-sg/src/config.ts
```diff
@@ -0,0 +1,27 @@
+export type PipelineConfig = {
```
daniel-va, line 1: This can be an interface.

daniel-va (changes requested):

apps/server-asset-sg/docker/Dockerfile
vej-ananas: This is also called in `RUN npx nx build server-asset-sg --configuration=production` which happens in the next line, see https://github.com/swisstopo/swissgeol-assets-suite/blob/ef5764841053e4f4f6d7f464acd7e27ccce9e234/apps/server-asset-sg/project.json#L14

PR #272, opened 2024-09-17
Retain detail with polygons

### LINKED ISSUE #240: Bug: Suche und URL out of Sync

vej-ananas:
resolves #240

### DIFF
--- a/libs/asset-viewer/src/lib/state/asset-search/asset-search.reducer.ts
+++ b/libs/asset-viewer/src/lib/state/asset-search/asset-search.reducer.ts
@@ -85,6 +85,7 @@ export const assetSearchReducer = createReducer(
     (state): AssetSearchState => ({
       ...state,
       resultsLoadingState: LoadingState.Loading,
+      filterLoadingState: LoadingState.Loading,
     })
   ),
   on(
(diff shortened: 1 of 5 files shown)


libs/asset-viewer/src/lib/state/asset-search/asset-search.reducer.ts
```diff
@@ -78,6 +78,12 @@ export const assetSearchReducer = createReducer(
         ...state.query,
         ...query,
       },
+    })
+  ),
+  on(
+    actions.loadSearch,
```
TIL-EBP, line 84: Why do we need this? Is this because of the loading state?
(Currently we have three different loading states, but actually just need one, because the loading bar is displayed as soon as one state is set to loading, right?) --> Then setting the filterLoadingState on actions.resetSearch can be removed, because we already set the loadingState on loadSearch

TIL-EBP (approved):

TIL-EBP (approved):

PR #273, opened 2024-09-19
Prevent admins from editing in workgroups they are no members of

### LINKED ISSUE #252: Admins sollen Workgroup Regeln unterliegen
**Beschreibung**
Admins sollen den Workgroup Regeln unterliegen für Editor und MasterEditor Funktionalitäten.
Admins haben immer mindestens Viewer für alle Workgroups.

**Acceptance Criteria**

- [ ] Admins können alle Workgroups und deren Assets sehen, auch ohne Workgroup.Viewer Rolle
- [ ] Admins können nur Assets bearbeiten, wenn sie die entsprechende Workgroup.Rolle haben

vej-ananas:
resolves #252

### DIFF
--- a/libs/shared/v2/src/lib/policies/workgroup.policy.ts
+++ b/libs/shared/v2/src/lib/policies/workgroup.policy.ts
@@ -4,11 +4,11 @@ import { Policy } from './base/policy';
 export class WorkgroupPolicy extends Policy<Workgroup> {
   canShow(value: Workgroup): boolean {
     // A user can see every workgroup assigned to them.
-    return this.hasWorkgroup(value.id);
+    return this.user.isAdmin || this.hasWorkgroup(value.id);
   }
 
   canCreate(): boolean {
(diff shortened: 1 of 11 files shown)


libs/shared/v2/src/lib/policies/workgroup.policy.ts
```diff
@@ -9,6 +9,6 @@ export class WorkgroupPolicy extends Policy<Workgroup> {
 
   canCreate(): boolean {
     // Only admins can create workgroups.
-    return false;
+    return this.canDoEverything() || false;
```
daniel-va, line 12: I would fully get rid of the `canDoEverything` method, e.g. remove it from the abstract `Policy` class. There are some spots where it is still in use, and might probably need to be removed. Here, it can just be replaced with `this.user.isAdmin()`.

daniel-va (changes requested):

daniel-va (approved):

PR #276, opened 2024-09-27
activate cypress pipeline

### LINKED ISSUE #210: Elasticsearch Errors in Cypress Pipeline
**Beschreibung**

In der Cypress Pipeline tauchen Elasticsearch Fehler auf. Bereinige diese.

**Acceptance Criteria**

- [ ] Es tauchen keine Elasticsearch Fehler mehr auf in der Cypress Pipeline.

DmByK:
resolves #210


development/init/testing/dump_3.sql
daniel-va, line 1: I would prefer to rename this file to something like `static_entities.sql` (or similar), just so the name makes it clear that we don't have any protected or personal data in here.

development/init/elasticsearch/init.sh
```diff
@@ -27,8 +27,74 @@ init_index() {
 
   echo "Initializing index ${INDEX_NAME}..."
 
-  curl --silent -XPUT "${ELASTICSEARCH_URL}/${INDEX_NAME}"
-  curl --silent -XPUT "${ELASTICSEARCH_URL}/${INDEX_NAME}/_mapping" -d "@/init/mappings/${INDEX_NAME}.json"
```
daniel-va, line 31: I would prefer if we could load the mapping from the json file instead of copying it plain-text here, so we only need to modify the mapping in one place when we want to make changes to it. This this syntax to read the file via parameter not work for you?

e2e/cypress/e2e/common/create-test-asset.ts
```diff
@@ -3,32 +3,33 @@ import { Given } from '@badeball/cypress-cucumber-preprocessor';
...
+            method: 'POST',
             url: 'http://localhost:4200/api/asset-edit',
             auth: bearerAuth(token),
             body: body,
             headers: {
               'Content-Type': 'application/json',
             },
           }).then((response) => {
-            assert.equal(response.status, 200, 'Response status should be 200');
+            expect(response.status).to.match(/^2\d{2}$/);
```
daniel-va, line 32: You can probably just match `201` - `Created` instead of allowing any 2XX code.

e2e/cypress/e2e/create-asset/create-asset.ts
```diff
@@ -60,7 +65,7 @@ Then(/^The user should see the Create Asset form$/, () => {
 Then(/^The asset is created$/, () => {
   cy.wait('@save')
     .then((xhr) => {
-      expect(xhr.response.statusCode).to.eq(200);
+      expect(xhr.response.statusCode).to.match(/^2\d{2}$/);
```
daniel-va, line 68: Same here as above, I prefer to see `201` over a `2XX` regexp.

daniel-va (changes requested):

PR #281, opened 2024-10-01
Feature 28: Manage Legal Docs

### LINKED ISSUE #28: Store, manage and display legal docs separately from the main asset files
Currently, only one type of asset-file relationship is implemented in the application. This relationship is used both for the actual geological documents, as well as for the legal consent documents. In the database schema, this relationship is represented in the `asset_file` table, which only has two columns: `asset_id` (a foreign key to the `asset` table) and `file_id` (a foreign key to the `file` table). This has several disadvantages:

* Ideally, we would like the legal docs to be accessible/visible only to a restricted set of (internal) users.
* In the user interface ("Detailinformationen" for a specific asset), the legal docs should be displayed differently and less prominently compared to the actual content, so that the user can identify more easily which file is the one they are interested in.
* Automatic data processing pipelines currently have no way of knowing for sure which files should or shouldn't be considered for tasks like classification or language identification (the legal docs should generally be ignored for such tasks).

image

The database schema currently already contains tables `legal_doc` (with columns `legal_doc_id`, `asset_id`, `title` and `legal_doc_item_code`; empty as of September 2023) and `legal_doc_item` (with codes "contract", "federalData", "other" and "permissionForm" defined as of September 2023). These tables are not used anywhere in the application. It seems like these tables were already added to the database schema in anticipation of a feature similar to the one described in this ticket. However, money probably ran out before the feature was actually implemented.

**ACs**

* Database schema migration
  * Add an optional/nullable column `file_id` (with a foreign key that refers to the `file` table) to the `legal_doc` table
* Database data migration
  * All existing files with a `file_name` that ends with "_LDoc.pdf" should be classified as legal docs.
    * Add a new entry to the table `legal_doc` with
      * `asset_id` → read from the `asset_file` table
      * `file_id` → read from the `file` table
      * `title` → empty string
      * `legal_doc_item_code` → "permissionForm"
    * Remove the corresponding rows in "asset_file" for these files whose filename ends with "_LDoc.pdf"
* Application: view detailed information of asset
  *  The legal docs (if any) should be displayed differently and less prominently compared to the main asset files.
  *  Suggestion:
     * Add a new row "Legal consent" / "Rechtliche Einwilligungen" in the table, between the existing rows "Kontakte" and "Thema".
     * For every legal doc that is linked to the current asset, show a line with `<legal doc type>: [<optional title>] [<filename> (<filesize>) <open in new tab icon> <download icon>]`.
       * For example: `Permission form: a1004_1003_LDoc.pdf (1MB) <open icon> <download icon>`
       * The part `<filename> (<filesize>) <open in new tab icon> <download icon>` can be reused from how the files / legal docs are currently displayed in the "File"/"Datei" row at the top of the table.
     * In the "File"/"Datei" row at the top of the "detailed information" table, only show those files that are still linked through the `asset_file` table (i.e. the legal docs are no longer shown here).
* Application: create/edit asset
  * Enable uploading/removing legal documents separately from the main asset files.
  * It should be possible to specify any number (0-unlimited) of legal docs for each asset.
  * Enable selecting the type ("contract", "federalData", "other" or "permissionForm") of each legal doc. This value must be set.
  * Enable specifying an optional title for the legal doc. This value can be empty.
  * Enable adding a legal doc without uploading any file. (This will be used primarily for the "federalData" type, but this should not be implemented as a formal constraint / validation rule.)

daniel-va:
Resolves #28.

Extends the `file` table to support legal documents.


README.md
```diff
@@ -12,32 +12,30 @@ The following components must be installed on the development computer:
...
 Generate prisma-client for database-access:
 
 ```bash
 npm run prisma -- generate
 ```
 
-#### 4. Initialize MinIO
+#### 3. Initialize MinIO
+
+> Note that this skip may be skipped if you do not need to interact with uploaded files,
```
vej-ananas, line 37:
```suggestion
> Note that this step may be skipped if you do not need to interact with uploaded files,
```

apps/server-asset-sg/prisma/migrations/20241001062933_add_legal_files/migration.sql
```diff
@@ -0,0 +1,37 @@
...
+ALTER TABLE "asset" DROP CONSTRAINT "asset_workgroup_id_fkey";
+ALTER TABLE "asset" ADD CONSTRAINT "asset_workgroup_id_fkey" FOREIGN KEY ("workgroup_id") REFERENCES "workgroup"("id") ON DELETE RESTRICT ON UPDATE CASCADE;
+ALTER TABLE "asset" ALTER COLUMN "workgroup_id" SET NOT NULL;
+
+-- CreateEnum
+CREATE TYPE "FileType" AS ENUM ('normal', 'legal');
+
+-- AlterTable
+ALTER TABLE "file" RENAME COLUMN "file_id" TO "id";
+ALTER TABLE "file" RENAME COLUMN "file_date" TO "last_modified_at";
```
vej-ananas, line 11: is there really a concept of modifiying a file? isnt there just the upload/creation date?

apps/server-asset-sg/src/features/files/file.repo.ts
vej-ananas, line 1: use the same naming for repo and controller (file.xyz)

apps/server-asset-sg/src/features/files/file.repo.ts
```diff
@@ -0,0 +1,155 @@
...
+      id,
+      name,
+      size: data.size,
+      type: data.type,
+      legalDocItemCode: data.legalDocItemCode,
+    };
+  }
+
+  private async determineUniqueFilename(fileName: string, assetId: AssetId): Promise<string> {
+    const name = fileName.startsWith('a' + assetId + '_') ? fileName : 'a' + assetId + '_' + fileName;
```
vej-ananas, line 55: These name creation rules look like a perfect unit test target to me

vej-ananas, line 55: These name creation rules look like a perfect unit test target to me

apps/server-asset-sg/src/features/files/files.controller.ts
```diff
@@ -57,41 +65,73 @@ export class FilesController {
...
+    }
+
+    const legalDocItemCode = pipe(
+      D.nullable(LegalDocItemCode).decode((req.body as { legalDocItemCode?: string }).legalDocItemCode ?? null),
+      E.getOrElseW(() => false as const)
+    );
+    if (legalDocItemCode === false) {
+      throw new HttpException('invalid legalDocItemCode', HttpStatus.BAD_REQUEST);
+    }
+    switch (type) {
```
vej-ananas, line 95: These kind of rules look like standard validation for me - we have the package 'class-validator' which is also recommended by nestjs, is this applicable here?

libs/asset-editor/src/lib/components/asset-editor-tab-general/asset-editor-tab-general.component.html
```diff
@@ -163,42 +163,70 @@
...
+                <ng-container *ngIf="file.willBeDeleted; else fileFormFields">
+                  <div *ngIf="file.willBeDeleted" class="notice text-sm text-grey-08" translate>
+                    edit.tabs.general.willBeDeleted
+                  </div>
+                </ng-container>
+
+                <ng-template #fileFormFields>
+                  <div class="type">
+                    <mat-select [value]="file.type" (valueChange)="updateFileType(file, $event)">
+                      <mat-option class="ellipsis" value="Normal">
```
vej-ananas, line 185: The legal docs are to be shown and edited at a different place than normal files, see issue - so we dont need a select like this because we implicitly always know the type

libs/asset-viewer/src/lib/components/asset-viewer-files/asset-viewer-files.component.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  @Input({ required: true })
+  type!: AssetFileType;
+
+  private readonly store = inject(Store<AppState>);
+
+  private readonly httpClient = inject(HttpClient);
+
+  public readonly legalDocItems$: Observable<Record<LegalDocItemCode, TranslatedValue>> = this.store
+    .select(fromAppShared.selectRDReferenceData)
```
vej-ananas, line 32: we typically do all of these in a separate selector - then this component does little more than display the finished files including correct translations and locale. 
I imagine there already exists such a selector for the AssetDetail, if you just add it there then the input to this component has everything

vej-ananas (changes requested):

apps/server-asset-sg/prisma/migrations/20241001062933_add_legal_files/migration.sql
daniel-va: This is just the file's modification date, and we never change it after the initial upload. The field is never used, so IMO we can also remove it.

apps/server-asset-sg/src/features/files/files.controller.ts
daniel-va: Yes and no - `class-validator` could validate this. However, the asset-files are part of the `AssetDetail`/`AssetEdit` types, which are the only part of Assets that is still using `fp-ts`/`io-ts` for validation. So to check these correctly, we would have to move all asset things to `class-validator`, which we have so far avoided.

For now, I can rewrite the `FilesController` to use `class-validator`, but everywhere else, we're stuck with the current types.

vej-ananas:
Also, please add an entry to the changelog

apps/server-asset-sg/src/features/files/file.repo.ts
daniel-va: All other features follow the same rule that is used here (controller -> plural, everything else -> singular). I wouldn't say that this is perfect (although I can explain why I named them this way :P). If we break this convention here, we should plan to do the same everywhere else, too.

vej-ananas: lol.. ich just checked asset-edit :)

vej-ananas (approved):

PR #311, opened 2024-10-21
Fix: don't allow to update usage status or work status if not masterEditor

### LINKED ISSUE #309: Bug:  Admins sollen keine MasterEditor Berechtigungen haben, wenn sie nur Editor sind
**Beschreibung**
Derzeit haben Administratoren auch dann MasterEditor Berechtigungen, wenn sie in einer Workgroup lediglich Editor sind.

Das macht sich folgend bemerkbar:
- in der Rolle "Editor" im Reiter "Nutzung" kann man die Externe Nutzung bearbeiten und speichern, nicht aber die Interne Nutzung (da wird eine Fehlermeldung ausgegeben). Eine Bearbeitung der Nutzung sollte in der Editor-Rolle nicht möglich sein.
- in der Rolle "Editor" im Reiter "Administration" kann man den Arbeitsstatus auf "Publiziert" stellen, was in der Rolle Editor ebenfalls nicht möglich sein sollte.

**Erwartetes Verhalten**
Ein Administrator ist für das Bearbeiten an die bestehenden Rollenkonzepte gebunden.

TIL-EBP:
resolves #309

### DIFF
--- a/apps/server-asset-sg/src/features/asset-edit/asset-edit.controller.ts
+++ b/apps/server-asset-sg/src/features/asset-edit/asset-edit.controller.ts
@@ -68,52 +68,27 @@ const validatePatch = (user: User, patch: PatchAsset, record?: AssetEditDetail) 
     );
   }
 
-  const checkStatusChange = (
-    hasChanged: boolean,
-    newStatus: string,
-    allowedStatus: string,
-    errorMessage: string,
-    record: AssetEditDetail | undefined,
-    policy: AssetEditPolicy,
-    patchWorkgroupId: number
-  ) => {
-    if (
+  const checkStatusChange = (key: 'internalUse' | 'publicUse', newStatus: string) => {
+    const hasChanged = record == null || record[key].statusAssetUseItemCode !== patch[key].statusAssetUseItemCode;
+    return (
       hasChanged &&
-      newStatus !== allowedStatus &&
+      newStatus !== 'tobechecked' &&
       ((record != null && !policy.hasRole(Role.MasterEditor, record.workgroupId)) ||
-        !policy.hasRole(Role.MasterEditor, patchWorkgroupId))
-    ) {
-      throw new HttpException(errorMessage, HttpStatus.FORBIDDEN);
-    }
+        !policy.hasRole(Role.MasterEditor, patch.workgroupId))
+    );
   };
 
   // Specialization of the policy where we disallow the internal status to be changed to anything else than `tobechecked`
   // if the current user is not a master-editor for the asset's current or future workgroup.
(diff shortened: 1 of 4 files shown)


apps/server-asset-sg/src/features/asset-edit/asset-edit.controller.ts
```diff
@@ -69,16 +68,60 @@ const validatePatch = (user: User, patch: PatchAsset, record?: AssetEditDetail)
...
+  ) => {
+    if (
+      hasChanged &&
+      newStatus !== allowedStatus &&
+      ((record != null && !policy.hasRole(Role.MasterEditor, record.workgroupId)) ||
+        !policy.hasRole(Role.MasterEditor, patchWorkgroupId))
+    ) {
+      throw new HttpException(errorMessage, HttpStatus.FORBIDDEN);
+    }
+  };
```
daniel-va, line 88: I have some issues with this function. Note that these are just general observations - if and how you apply them is up to you. Solving some of these issues even makes others obsolete.

- There is currently no reason for this to be a nested function (a function within another function). I like nested functions, as long as they reuse their outer scope, making their parameter list more readable. Here, the last three parameters are unnecessary - they can be replaced by accessing the outer function's scope.

- As a general rule, I like to keep a function's parameter count at a maximum of 3 or 4. Any more than that and it becomes unclear what each argument passed to the function does. If there is no way to reduce the parameter count, use an options object, e.g:
  ```ts
  const checkStatusChange = (hasChanged: boolean, options: {
    newStatus: string;
    allowedStatus: string;
    errorMessage: string;
  }) => { ... } 
  ```
  This makes function calls much more readable:
  ```ts
  checkStatusChange(hasInternalUseChanged, {
    newStatus: patch.internalUse.statusAssetUseItemCode,
    allowedStatus: 'tobechecked',
    errorMessage: 'Changing the asset's status is not allowed',
  })
  ```

- There is no reason for `allowedStatus` and `errorMessage` to be parameters, as they are the same for all function calls.

- Instead of passing `hasChanged` and `newStatus`, you could just pass the key `'internalUse' | 'publicUse'`. This would also enable you to compute `hasChanged` within the function instead of duplicating it outside of it.

- Reading a function starting with `check`, I would expect it to return a value (a `boolean`, most likely). Here, I would prefer `validateStatus` to keep with the outer function's name, or even `validateStatusOrThrow` if you want to be explicit about what it does.
If you want to keep it as a `check*`, I personally would prefer to make the function return a `boolean` and then make the function caller throw the exception themself. This would also enable you to remove the `errorMessage` parameter from the function, while keeping the function pure.

daniel-va (approved):

PR #313, opened 2024-10-22
Feature 70: add disclaimer

### LINKED ISSUE #70: Swissgeol Disclaimer (Assets)
**Beschreibung**
Es muss ein Disclaimer gezeigt werden, vor die Applikation genutzt werden kann. 
Der Disclaimer muss akzeptiert werden. Er tauch nach jedem Reload der Applikation erneut auf.

Das Design soll sich an Boreholes anlehnen
Image

**EN: Terms of service**
Limitation of liability
Although every care has been taken by the Federal Office of Topography swisstopo to ensure the accuracy of the information published, no guarantee can be given with regard to the accurate, reliable, up-to-date or complete nature of this information.

swisstopo reserves the right to alter or remove the content, in full or in part, without prior notice.

Liability claims against swisstopo for material or immaterial damage resulting from access to or use or non-use of the published information, from misuse of the connection or from technical faults are excluded.

Further legal provisions can be found here: https://www.swissgeol.ch/datenschutz-en/

**DE: Nutzungsbedingungen**
Haftung
Obwohl das Bundesamt für Landestopografie swisstopo mit aller Sorgfalt auf die Richtigkeit der veröffentlichten Informationen achtet, kann hinsichtlich der inhaltlichen Richtigkeit, Genauigkeit, Aktualität, Zuverlässigkeit und Vollständigkeit dieser Informationen keine Gewährleistung übernommen werden.

Swisstopo behält sich ausdrücklich vor, jederzeit Inhalte ohne Ankündigung ganz oder teilweise zu ändern, zu löschen oder zeitweise nicht zu veröffentlichen.

Haftungsansprüche gegen swisstopo wegen Schäden materieller oder immaterieller Art, welche aus dem Zugriff oder der Nutzung bzw. Nichtnutzung der veröffentlichten Informationen, durch Missbrauch der Verbindung oder durch technische Störungen entstanden sind, werden ausgeschlossen.

Weitere rechtliche Bestimmungen finden sie hier: https://www.swissgeol.ch/datenschutz/

**FR: Conditions d'utilisation**
Responsabilité
Malgré la grande attention qu’il porte à la justesse des informations diffusées sur ce site, l’Office fédéral de topographie swisstopo ne peut endosser aucune responsabilité quant à la fidélité, à l’exactitude, à l’actualité, à la fiabilité et à l’intégralité de ces informations.

swisstopo se réserve expressément le droit de modifier en partie ou en totalité le contenu de ce site, de le supprimer ou d’en suspendre temporairement la diffusion, et ce à tout moment et sans avertissement préalable.

swisstopo ne saurait être tenu pour responsable des dommages matériels ou immatériels qui pourraient être causés par l’accès aux informations diffusées ou par leur utilisation ou non-utilisation, par le mauvais usage de la connexion ou par des problèmes techniques.

Pour plus d'informations légales, veuillez consulter le lien suivant: https://www.swissgeol.ch/datenschutz-fr/

**IT: Condizioni di utilizzo**
Responsabilità
Nonostante si presti grande attenzione all’esattezza delle informazioni pubblicate su questo sito, l’Ufficio federale di topografia swisstopo declina ogni responsabilità per la fedeltà, l’esattezza, l’attualità, l’affidabilità e la completezza di tali informazioni.

swisstopo si riserva esplicitamente il diritto in qualsiasi momento di modificare parzialmente o completamente il contenuto del sito, di cancellarlo o di sospenderne temporaneamente la pubblicazione, senza alcun preavviso.

swisstopo declina ogni responsabilità per danni materiali o immateriali derivanti dall’accesso alle informazioni diffuse, dall’uso o dal mancato uso del sito, oppure che sono riconducibili a un malfunzionamento del collegamento o a disturbi tecnici del sito.

Per ulteriori informazioni legali, consultare il seguente link: https://www.swissgeol.ch/datenschutz-it/

TIL-EBP:
resolves #70 

to test, set `hideDisclaimer `in `environment.ts` to `false`


libs/auth/src/lib/services/auth.service.ts
```diff
@@ -44,6 +47,10 @@ export class AuthService {
           // If something else has interrupted the auth process, then we don't want to signal a success.
           if (this._state$.value === AuthState.Ongoing) {
             this._state$.next(AuthState.Success);
+            this.dialogService.open<DisclaimerDialogComponent>(DisclaimerDialogComponent, {
```
daniel-va, line 50: The type parameter can be dropped here.
```suggestion
         this.dialogService.open(DisclaimerDialogComponent, {
```

apps/client-asset-sg/src/app/components/language-selector/language-selector.component.ts
```diff
@@ -0,0 +1,61 @@
...
+import { LetModule } from '@rx-angular/template/let';
+import { flow, pipe } from 'fp-ts/function';
+import * as O from 'fp-ts/Option';
+import queryString from 'query-string';
+import { debounceTime, filter, map, startWith } from 'rxjs';
+
+@UntilDestroy()
+@Component({
+  selector: 'asset-sg-language-selector',
+  standalone: true,
```
daniel-va, line 19: Instead of making this a standalone component, you can just move it to the `client-shared` module.

apps/client-asset-sg/src/app/components/language-selector/language-selector.component.ts
```diff
@@ -0,0 +1,61 @@
...
+import { Lang } from '@asset-sg/shared';
+import { SvgIconComponent } from '@ngneat/svg-icon';
+import { UntilDestroy } from '@ngneat/until-destroy';
+import { LetModule } from '@rx-angular/template/let';
+import { flow, pipe } from 'fp-ts/function';
+import * as O from 'fp-ts/Option';
+import queryString from 'query-string';
+import { debounceTime, filter, map, startWith } from 'rxjs';
+
+@UntilDestroy()
```
daniel-va, line 16: This doesn't do anything here as the component doesn't use any manually managed subscriptions.

apps/client-asset-sg/src/app/components/disclaimer-dialog/disclaimer-dialog.component.ts
```diff
@@ -0,0 +1,72 @@
...
+    MatDivider,
+    LanguageSelectorComponent,
+    NewlineToBrPipe,
+    LetModule,
+  ],
+  templateUrl: './disclaimer-dialog.component.html',
+  styleUrl: './disclaimer-dialog.component.scss',
+})
+export class DisclaimerDialogComponent {
+  constructor(private readonly dialogRef: MatDialogRef<DisclaimerDialogComponent>, private readonly router: Router) {}
```
daniel-va, line 47: These constructor parameters can be `inject`ed.

apps/client-asset-sg/src/app/components/language-selector/language-selector.component.ts
```diff
@@ -0,0 +1,61 @@
...
+      supportedLangs.map((lang) => ({
+        isActive: lang === currentLang.lang,
+        lang: lang.toUpperCase(),
+        params: [`/${lang}${currentLang.path}`],
+        queryParams: currentLang.queryParams,
+      }))
+    )
+  );
+
+  constructor(private readonly router: Router) {}
```
daniel-va, line 60: These constructor parameter can be `inject`ed.

apps/client-asset-sg/src/app/components/disclaimer-dialog/disclaimer-dialog.component.ts
```diff
@@ -0,0 +1,72 @@
...
+
+  public readonly href = 'https://www.swissgeol.ch/datenschutz';
+  public readonly currentLang$ = this.router.events.pipe(
```
daniel-va, line 50: This isn't the current language - I would replace this with something like `legalUrl`.

apps/client-asset-sg/src/app/components/disclaimer-dialog/disclaimer-dialog.component.ts
```diff
@@ -0,0 +1,72 @@
...
+    NewlineToBrPipe,
+    LetModule,
+  ],
+  templateUrl: './disclaimer-dialog.component.html',
+  styleUrl: './disclaimer-dialog.component.scss',
+})
+export class DisclaimerDialogComponent {
+  constructor(private readonly dialogRef: MatDialogRef<DisclaimerDialogComponent>, private readonly router: Router) {}
+
+  public readonly href = 'https://www.swissgeol.ch/datenschutz';
```
daniel-va, line 49: This doesn't need to be public and is constant - I would move this to something like `const LEGAL_BASE_URL` outside of the component class.

apps/client-asset-sg/src/app/components/disclaimer-dialog/disclaimer-dialog.component.ts
```diff
@@ -0,0 +1,72 @@
...
+import { flow, pipe } from 'fp-ts/function';
+import * as O from 'fp-ts/Option';
+import queryString from 'query-string';
+import { filter, map, startWith } from 'rxjs';
+import { NewlineToBrPipe } from '../../../../../../libs/client-shared/src/lib/pipes/new-lline-to-br.pipe';
+import { LanguageSelectorComponent } from '../language-selector/language-selector.component';
+
+@Component({
+  selector: 'asset-sg-disclaimer-dialog',
+  standalone: true,
```
daniel-va, line 24: This doesn't need to be a standalone component.

apps/client-asset-sg/src/app/components/disclaimer-dialog/disclaimer-dialog.component.html
```diff
@@ -0,0 +1,16 @@
+<div>
+  <div class="header">
+    <h2 translate>disclaimer.title</h2>
+    <asset-sg-language-selector></asset-sg-language-selector>
+  </div>
+  <mat-divider></mat-divider>
+  <div mat-dialog-content>
+    <h1 translate>disclaimer.subtitle</h1>
+    <p [innerHTML]="'disclaimer.content' | translate | newlineToBr"></p>
+    <a *rxLet="currentLang$; let lang" [href]="lang" target="_blank" rel="noopener noreferrer">{{ lang }}</a>
```
daniel-va, line 10: `noreferrer` is debatable here (it's more than okay to let the swissgeol main page know that we send users to it). I would also definitely add `external`, and maybe even `nofollow`.

apps/client-asset-sg/src/app/components/app-bar/app-bar.component.html
```diff
@@ -5,31 +5,7 @@
...
-        <svg-icon key="checkmark" />
-      </span>
-      <span class="name">
-        {{ language.lang }}
-      </span>
-    </a>
-    }
-  </ng-container>
-</mat-menu>
+<asset-sg-language-selector></asset-sg-language-selector>
```
daniel-va, line 8: This can be shortened.
```suggestion
<asset-sg-language-selector />
```

apps/client-asset-sg/src/app/i18n/en.ts
```diff
@@ -285,4 +285,17 @@ export const enAppTranslations: AppTranslations = {
       addUsers: 'Add users',
     },
   },
+  disclaimer: {
+    title: 'Terms of service',
+    subtitle: 'Limitation of liability',
```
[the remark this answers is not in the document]
TIL-EBP: I copied this from the ticket and boreholes also has it lowercase, so I would just leave it at that unless they correct it themselves

libs/client-shared/src/lib/pipes/new-lline-to-br.pipe.ts
```diff
@@ -0,0 +1,11 @@
+import { Pipe, PipeTransform } from '@angular/core';
+
+@Pipe({
+  standalone: true,
+  name: 'newlineToBr',
+})
+export class NewlineToBrPipe implements PipeTransform {
```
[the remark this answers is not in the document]
TIL-EBP: Do you mean defining the pipe inside the component file or not using a pipe at all?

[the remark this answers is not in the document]
daniel-va: I'd not use a pipe at all.

apps/client-asset-sg/src/app/components/disclaimer-dialog/disclaimer-dialog.component.scss
```diff
@@ -1,22 +1,34 @@
+::ng-deep .cdk-overlay-dark-backdrop {
+  background-color: grey;
+}
```
daniel-va, line 3: `::ng-deep` **always** needs a prefix that is not inside the `::ng-deep`, as it is otherwise makes everything after it global.
```suggestion
:host ::ng-deep .cdk-overlay-dark-backdrop {
  background-color: grey;
}
```
> Note that `::ng-deep` is [deprecated](https://stackoverflow.com/a/49308475), without adding a replacement. This means there is no other choice but to use it in situations like these, while knowing that it is has some issues and might be removed at some point.

apps/client-asset-sg/src/app/components/disclaimer-dialog/disclaimer-dialog.component.ts
```diff
@@ -1,69 +1,27 @@
...
-        O.bind('parsed', ({ match }) => O.of(queryString.parseUrl(match[2]))),
-        O.map(({ lang }) => lang)
-      )
-    ),
-    map((it) => O.toNullable(it)),
-    filter(isNotNull),
-    map((lang) => (lang === 'de' ? this.href : `${this.href}-${lang}/`))
+  public readonly disclaimerText$ = getCurrentLang().pipe(
+    map(() => this.translateService.instant(`disclaimer.content`).replace(/\n/g, '<br>'))
   );
```
daniel-va: 1. You can `TranslateService.get` over `TranslateService.instant` to get an observable of the translated value that emits if the language changes. This allows you to skip using `getCurrentLang` as base observable.
2. You can replace the regex replacement with a `replaceAll` using the newline character.
```suggestion
  public readonly disclaimerText$ = this.translateService.get(`disclaimer.content`).pipe(
    map((value: string) => value.replaceAll('\n', '<br>'))
  )
```

libs/client-shared/src/lib/utils/current-lang.ts
```diff
@@ -14,3 +18,28 @@ export function currentLangFactory(): Observable<string> {
...
+          lang,
+          path: parsed.url,
+          queryParams: parsed.query,
+        }))
+      )
+    ),
+    map((it) => O.toNullable(it)),
+    filter(isNotNull)
+  );
+}
```
daniel-va, line 45: I like the idea of this function, but I have some issues with it:

1. For every call to `getCurrentLang()`, we create a fully independent observable that listens to router events and processes them. Ideally, the observable is created only once, and then shared amongst all who need it. To do this, either place the observable in a service or store it in an [Injection Token](https://v17.angular.io/api/core/InjectionToken), and pipe its result into a [shareReplay](https://www.learnrxjs.io/learn-rxjs/operators/multicasting/sharereplay).
2. `path` can be typed with `string`, `queryParams` with `ParsedQuery<string>`.
3. The function's name reads as if the observable just returns the `lang` - I would rename it to something like `getParsedUrl`.

daniel-va (changes requested):

apps/client-asset-sg/src/app/components/disclaimer-dialog/disclaimer-dialog.component.ts
TIL-EBP: I get an error saying replaceAll does not exist on type String. When follwoing [this](https://stackoverflow.com/questions/63616486/property-replaceall-does-not-exist-on-type-string), a bunch of other error messages appear

libs/client-shared/src/lib/utils/current-lang.ts
TIL-EBP: The initial idea of this function was to reduce code duplication as the same code was used in the` language-selector` component and the `discalimer-dialog`. I now realised that there already is an INjectionToken for the current lang, which we can use in the dialog. This removes the need to have the `getCurrentLang()` (newly `getParsedUrl()`) method stored in a separate file and cna just be moved back to the `language-selector`.

apps/client-asset-sg/src/styles.scss
```diff
@@ -133,12 +143,18 @@ input {
   position: relative;
 }
 
+.disclaimer-backdrop-color {
+  background-color: grey;
+}
```
daniel-va, line 148: Move this back into the disclaimer CSS (using `::ng-deep`) - Its not a problem to have "global" css in the component modules, as long as it is clean (i.e. as safe as possible) and commented as such.

daniel-va (approved):

daniel-va:
# Squash Commits before Merge!

PR #340, opened 2024-11-07
Hotfix: Clear Selected Asset in Search

daniel-va:
Rewrites the search effects to fix any remaining errors with the search parameters and to make handling the entire search easier.

Behaviour should not be changed from DEV.

### DIFF
--- a/libs/asset-viewer/src/lib/state/asset-search/asset-search.effects.ts
+++ b/libs/asset-viewer/src/lib/state/asset-search/asset-search.effects.ts
@@ -1,24 +1,24 @@
 import { inject, Injectable } from '@angular/core';
 import { ActivatedRoute, Params, Router } from '@angular/router';
-import { AppState, assetsPageMatcher, fromAppShared } from '@asset-sg/client-shared';
-import { deepEqual, isDecodeError, isNotNull, ORD } from '@asset-sg/core';
-import { AssetSearchQuery, AssetSearchResult, LV95, Polygon } from '@asset-sg/shared';
-import * as RD from '@devexperts/remote-data-ts';
+import { AppState } from '@asset-sg/client-shared';
+import { isNotNull, ORD, isNull } from '@asset-sg/core';
+import { AssetSearchQuery, LV95, Polygon } from '@asset-sg/shared';
+import { AuthService } from '@asset-sg/auth';
 import { UntilDestroy } from '@ngneat/until-destroy';
-import { Actions, concatLatestFrom, createEffect, ofType } from '@ngrx/effects';
-import { ROUTER_NAVIGATED } from '@ngrx/router-store';
+import { Actions, createEffect, ofType } from '@ngrx/effects';
 import { Store } from '@ngrx/store';
-import * as D from 'io-ts/Decoder';
-import { EMPTY, filter, map, merge, of, shareReplay, switchMap, takeUntil, withLatestFrom } from 'rxjs';
+import { combineLatest, filter, map, of, switchMap, take, withLatestFrom } from 'rxjs';
 
 import { AllStudyService } from '../../services/all-study.service';
 import { AssetSearchService } from '../../services/asset-search.service';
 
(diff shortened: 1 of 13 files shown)


TIL-EBP (changes requested):
I havent looked at the code in detail but I found these two things whliplaying with the ui: 
- The filters are no longer written to the url
- On reload, the selected asset is loaded again but removed form the url. This means it is lost on a second reload

libs/asset-viewer/src/lib/state/asset-search/asset-search.effects.ts
```diff
@@ -27,208 +26,160 @@ import {
...
+        ];
+      })
     )
   );
 
-  public assetResults$ = createEffect(() =>
+  public updateParamsFromQuery$ = createEffect(
+    () =>
+      combineLatest([
+        this.store.select(selectAssetSearchQuery),
```
TIL-EBP, line 66: Even though this is valid, I prefer effects to be triggered by actions instead of state changes, as is intended by ngRx

libs/asset-viewer/src/lib/state/asset-search/asset-search.effects.ts
```diff
@@ -27,208 +26,160 @@ import {
...
-    );
-  });
-
   public loadStudies$ = createEffect(() => {
     return this.actions$.pipe(
-      ofType(actions.initializeSearch),
-      switchMap(() => this.store.select(selectStudies)),
-      filter((x) => !x),
+      ofType(actions.initialize),
+      switchMap(() => this.store.select(selectStudies).pipe(take(1))),
```
TIL-EBP, line 120: take(1) is not needed, right? The effect only triggers on actions.initialize, and this is only triggered once

TIL-EBP (changes requested):

libs/asset-viewer/src/lib/state/asset-search/asset-search.effects.ts
```diff
@@ -27,208 +23,187 @@ import {
...
     this.actions$.pipe(
-      ofType(actions.loadSearch),
-      switchMap(({ query }) =>
-        this.assetSearchService
-          .search(query)
-          .pipe(map((searchResults: AssetSearchResult) => actions.updateSearchResults({ searchResults })))
-      )
+      ofType(actions.mergeQuery),
+      withLatestFrom(this.store.select(selectAssetSearchQuery)),
+      map(([{ query }, currentQuery]) => actions.search({ query: { ...currentQuery, ...query } }))
```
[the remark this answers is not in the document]
daniel-va: You mean just use different names? Otherwise, your code is equivalent to mine.

libs/asset-viewer/src/lib/state/asset-search/asset-search.effects.ts
daniel-va: I'd argue it's in our best interest to make an exception here, as we want to always keep the params up-to-date with the query. Without reacting to state changes, this becomes much more cumbersome to track and quite easy to break.

libs/asset-viewer/src/lib/state/asset-search/asset-search.effects.ts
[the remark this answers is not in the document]
TIL-EBP: Exactly. It is not really important, I just sumbled across this and had to check twice which one is the currentQuery

libs/asset-viewer/src/lib/state/asset-search/asset-search.effects.ts
TIL-EBP: True. But I would say the NgRx way to do this would be to trigger this effect through any of the actions that change one of the required states. I do see, however, that this is more difficult to read, so it is fine by me

TIL-EBP (dismissed):

TIL-EBP (approved):

PR #357, opened 2024-11-25
Bugfix 356: Studies verschwinden, wenn man ein Detail anwählt

### LINKED ISSUE #356: Bug: Studies verschwinden, wenn man ein Detail anwählt
**Beschreibung**
Wenn man in die Heatmap reinzoomt und auf einen blauen Punkt klickt, dann erscheint die Detailansicht.
Bisher sind die restlichen blauen Punkte durchsichtig geworden, aber weiterhin sichtbar und klickbar.
Jetzt verschwinden sie stattdessen.
Das ursprüngliche Verhalten soll wiederhergestellt werden.

daniel-va:
Resolves #356.

### DIFF
--- a/libs/asset-viewer/src/lib/state/asset-search/asset-search.selector.ts
+++ b/libs/asset-viewer/src/lib/state/asset-search/asset-search.selector.ts
@@ -60,11 +60,15 @@ export const selectAssetSearchTotalResults = createSelector(assetSearchFeature, 
 export const selectCurrentAssetDetail = createSelector(assetSearchFeature, (state) => state.currentAsset);
 export const selectStudies = createSelector(assetSearchFeature, (state) => state.studies);
 
-export const selectHasDefaultFilters = createSelector(
+export const selectHasNoActiveFilters = createSelector(assetSearchFeature, ({ query }) => hasNoActiveFilters(query));
+
+export const selectIsSearchQueryEmpty = createSelector(
   assetSearchFeature,
-  ({ query, currentAsset }) =>
-    currentAsset == null && Object.values(query).every((value) => value === undefined || value == false)
+  ({ query, currentAsset }) => currentAsset == null && hasNoActiveFilters(query)
 );
+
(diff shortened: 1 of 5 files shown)


libs/asset-viewer/src/lib/state/asset-search/asset-search.selector.ts
```diff
@@ -60,12 +60,16 @@ export const selectAssetSearchTotalResults = createSelector(assetSearchFeature,
...
+export const selectHasDefaultFilters = createSelector(assetSearchFeature, ({ query }) => hasDefaultFilters(query));
+
+export const selectIsDefaultFilterState = createSelector(
   assetSearchFeature,
-  ({ query, currentAsset }) =>
-    currentAsset == null && Object.values(query).every((value) => value === undefined || value == false)
+  ({ query, currentAsset }) => currentAsset == null && hasDefaultFilters(query)
 );
 
+const hasDefaultFilters = (query: AssetSearchQuery): boolean =>
```
TIL-EBP, line 70: I am still not fully happy with the naming of these. 
I would prefer something like "isQueryEmpty" or "hasNoActiveFilters". 
And for `selectIsDefaultFilterState` i would suggest something like `isInitialSearchState`. I feel like this more accurately represents the status of the state

TIL-EBP (approved):

libs/asset-viewer/src/lib/state/asset-search/asset-search.selector.ts
daniel-va: Agreed. I changed the selectors to: 
- `selectHasDefaultFilters` -> `selectHasNoActiveFilters`
- `selectIsDefaultFilterState` -> `selectIsSearchQueryEmpty`

For the second one, I didn't want to use something like `isInitialSearchState`, as that may implicate that the search has not been executed yet. `selectIsSearchQueryEmpty` makes sense to me here in that neither the search nor a specific assets are set.

PR #376, opened 2025-01-20
Feature 335: Bug - Alternativ Ids

### LINKED ISSUE #335: Bug:  Alternativ Ids
**Beschreibung**
Das Verhalten beim editieren von Alternativ-Ids ist inkonsistent und unbenutzbar.

1. Gehe zur Seite Edit Asset
2. Stelle sicher, dass es 2 Alternativ-Ids gibt
3. Editiere den ersten Eintrag
4. Lösche den Text der Beschreibung
5. Die andere Alternativ-Id verschwindet. 
6. Auch beim Benutzen des Abbrechen Buttons verschwinden Alternativ-Ids

**Beobachtetes Verhalten**
Alternativ-Ids verschwinden ungewollt beim Editieren

**Erwartetes Verhalten**
Keine Alternativ-Id sollte gegen den eigenen Willen verschwinden.
Bei einer ungespeicherten Änderung im Editier-Dialog soll auch die Gesamtspeicherung verhindert werden.

daniel-va:
Resolves #335.

This is basically just a complete rewrite of the ids section in the asset form.
The section was really convoluted and basically not worth to fix as-is.
Now, ids are displayed and edited in two separate components, which should be much easier to understand and change in the future.

### DIFF
--- a/libs/asset-editor/src/lib/components/asset-editor-id-form/asset-editor-id-form.component.html
+++ b/libs/asset-editor/src/lib/components/asset-editor-id-form/asset-editor-id-form.component.html
@@ -1,0 +1,18 @@
+<form [formGroup]="form">
+  <mat-form-field class="mb-2">
+    <mat-label translate>edit.tabs.general.alternativeId</mat-label>
+    <input matInput formControlName="value" [required]="!isStandalone" />
+    <mat-error *ngIf="form.controls.value.hasError('required')" translate>required</mat-error>
+  </mat-form-field>
+  <mat-form-field class="mb-2">
+    <mat-label translate>edit.tabs.general.alternativeIdDescription</mat-label>
+    <input matInput formControlName="description" required="true" />
+    <mat-error *ngIf="form.controls.description.hasError('required')" translate>required</mat-error>
+  </mat-form-field>
+  <div class="self-end" *ngIf="!isStandalone">
+    <button asset-sg-primary class="mr-4" [disabled]="form.invalid" (click)="saveForm()">
+      {{ (isNew ? "add" : "save") | translate }}
+    </button>
+    <button asset-sg-secondary (click)="cancel()">Abbrechen</button>
+  </div>
+</form>
--- a/libs/asset-editor/src/lib/components/asset-editor-id-form/asset-editor-id-form.component.ts
+++ b/libs/asset-editor/src/lib/components/asset-editor-id-form/asset-editor-id-form.component.ts
@@ -1,0 +1,139 @@
+import { Component, EventEmitter, Input, OnChanges, OnDestroy, OnInit, Output, SimpleChanges } from '@angular/core';
+import { FormControl, FormGroup, Validators } from '@angular/forms';
+import * as O from 'fp-ts/Option';
+import { Subscription } from 'rxjs';
+import { eqIdVM, IdVM } from '../../models';
+
+@Component({
+  selector: 'asset-sg-asset-editor-id-form',
+  templateUrl: './asset-editor-id-form.component.html',
+  styleUrl: './asset-editor-id-form.component.scss',
+})
+export class AssetEditorIdFormComponent implements OnInit, OnChanges, OnDestroy {
+  @Input({ required: true })
+  public value!: IdVM | null;
+
+  @Input()
+  public isStandalone = false;
+
+  @Input()
+  public isDisabled = false;
+
+  @Output()
+  public save = new EventEmitter<IdVM>();
+
+  @Output()
+  public delete = new EventEmitter<IdVM>();
+
+  @Output()
+  public finish = new EventEmitter<void>();
+
+  private _valueAsVM: Partial<IdVM> | null = null;
+
+  public readonly form = new FormGroup({
+    id: new FormControl<number | null>(null),
+    value: new FormControl('', {
+      validators: [Validators.required],
+      nonNullable: true,
+      updateOn: 'blur',
+    }),
+    description: new FormControl('', { nonNullable: true }),
+  });
+
+  private readonly subscription = new Subscription();
+
+  public ngOnInit(): void {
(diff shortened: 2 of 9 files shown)


libs/asset-editor/src/lib/components/asset-editor-id-form/asset-editor-id-form.component.html
```diff
@@ -0,0 +1,18 @@
...
+    <mat-label translate>edit.tabs.general.alternativeId</mat-label>
+    <input matInput formControlName="value" [required]="!isStandalone" />
+    <mat-error *ngIf="form.controls.value.hasError('required')" translate>required</mat-error>
+  </mat-form-field>
+  <mat-form-field class="mb-2">
+    <mat-label translate>edit.tabs.general.alternativeIdDescription</mat-label>
+    <input matInput formControlName="description" required="true" />
+    <mat-error *ngIf="form.controls.description.hasError('required')" translate>required</mat-error>
+  </mat-form-field>
+  <div class="self-end" *ngIf="!isStandalone">
```
TIL-EBP, line 12: This does not work, becuase now, i cannot edit any alternate id (not sure if this only applies to assets that have just one alternate id)

libs/asset-editor/src/lib/components/asset-editor-id-list/asset-editor-id-list.component.html
```diff
@@ -0,0 +1,31 @@
...
+  </li>
+  }
+</ul>
+
+@if (isFormOpen) {
+<asset-sg-asset-editor-id-form [control]="control" [value]="formValue" (finish)="this.closeForm()" />
+} @if (!isFormOpen) {
+<button asset-sg-secondary class="mb-2 -mt-2 self-start" (click)="openForm()" translate>
+  <svg-icon key="zoom-plus" class="mr-4"></svg-icon>
+  edit.tabs.general.addNewAlternativeId
```
TIL-EBP, line 29: This never appears, thus i can no longer add alternate ids to exisiting assets

TIL-EBP: When creating a new asset, i can only add one alternate id, but i cannot save it, meaning it gets losts once i save the new asset

libs/asset-editor/src/lib/components/asset-editor-id-list/asset-editor-id-list.component.html
```diff
@@ -0,0 +1,31 @@
+<div class="font-bold mb-4 mt-2" translate>edit.tabs.general.alternativeId</div>
+
+<ul>
+  @for (id of ids; track id.id) {
+  <li class="bg-grey-00" [class.is-being-edited]="isFormValue(id)">
+    <span class="basis-full"> {{ id.id }} [{{ id.description }}] </span>
+    <button asset-sg-icon-button class="actions basis-10" [matMenuTriggerFor]="idListMenu">
+      <svg-icon key="action-menu"></svg-icon>
+    </button>
+    <mat-menu #idListMenu="matMenu">
```
TIL-EBP, line 10: This is not visible

libs/asset-editor/src/lib/components/asset-editor-id-list/asset-editor-id-list.component.html
```diff
@@ -0,0 +1,31 @@
+<div class="font-bold mb-4 mt-2" translate>edit.tabs.general.alternativeId</div>
+
+<ul>
+  @for (id of ids; track id.id) {
+  <li class="bg-grey-00" [class.is-being-edited]="isFormValue(id)">
+    <span class="basis-full"> {{ id.id }} [{{ id.description }}] </span>
+    <button asset-sg-icon-button class="actions basis-10" [matMenuTriggerFor]="idListMenu">
```
TIL-EBP, line 7: are we still using these classes? (basis-10)

TIL-EBP (changes requested):
I have not fully finished my review, because something is broken. I can no longe create new alternate ids or edit them (and actually save the edits). 
There seems to be something off with the conditions on when to display what. 

Also, I am not sure how much of this was refactored and how much was just moved to the new components, but now it seems like a mix between the old style and our new (and preferred style). If it is not too much effort I would suggest doing the refactoring all the way

libs/asset-editor/src/lib/components/asset-editor-id-form/asset-editor-id-form.component.ts
```diff
@@ -0,0 +1,139 @@
...
+      updateOn: 'blur',
+    }),
+    description: new FormControl('', { nonNullable: true }),
+  });
+
+  private readonly subscription = new Subscription();
+
+  public ngOnInit(): void {
+    // If this is the only persisted id, we write it to the parent form
+    // without requiring the save button to be pressed.
```
TIL-EBP, line 47: this does not quite work yet: 

edit an asset that has no alternative IDs yet
add a new one (there will be no save button)
click "Neue Alternativ-ID hinzufügen"
click hinzufügen --> the first id disappears

TIL-EBP: Edit: This goe for all Assets, that have just one alternate ID: If you try to add a new on that is empty, the original disappears

TIL-EBP: edit 2: same for cancelling: if there is one alternate ID, i try to create a new one and cancel, the original id disappears

TIL-EBP (changes requested):
It does work much better now, but i still found weird behavior when there is only one alterante id

stijnvermeeren (approved):
Test mit privaten Github Account

TIL-EBP (approved):

PR #406, opened 2025-02-05
Feature 386: Löschen von Asset führt zu Endlosschlaufe

### LINKED ISSUE #370: Bug: Link auf asset id funktioniert nicht
**Beschreibung**
Die URL `https://assets.swissgeol.ch/de?assetId=2163` sollte die Karte öffnen mit dem Asset 2163 in der Detailansicht und der Karte darauf gezoomt.
Das passiert nur, wenn es der erste Aufruf ist - vgl. Private Window. Bei wiederholtem Aufruf oder Reload bleibt die Ansicht auf der gesamten Karte ohne Detailansicht und zoom.

1. Öffne private window
2. Navigiere auf [https://assets.swissgeol.ch/de?assetId=2163](https://assets.swissgeol.ch/de?assetId=2163): Die Detailansicht öffnet sich und die Karte zoomt auf das Asset
3. Reloade die Seite
4. Die Karte bleibt auf der Gesamtansicht ohne Detail

**Erwartetes Verhalten**
Die URL führt konsistent zu derselben Ansicht auch nach Reload

### LINKED ISSUE #386: Löschen von Asset führt zu Endlosschlaufe
**Beschreibung:**
Löscht man aktuell ein Asset, gelangt die Applikation in eine "Endlosschlaufe" (vgl. Screenshot):

Image

Erst nach einem Reload der Applikation, einem Zurücksetzen der Suche oder einem Klick auf einen anderen Eintrag in der Trefferliste verschwindet der sich drehende Kreis auf der rechten Seite (Bereich der Detailansicht). Dies ist unschön und etwas irritierend.

**Schritte zum Reproduzieren:**
- Gehe zum Editiermodus eines Assets
- Lösche das Asset
- Die Applikation kehrt automatisch zur vorherigen Ansicht mit dem sich drehenden Kreis bei der Detailansicht

**Erwartetes Verhalten:**
Die Applikation zeigt danach die vorangehende Trefferliste (bei Auswahl mehrerer Assets) oder die Heatmap (bei vorheriger Anzeige des einzelnen, nun gelöschten Assets).

**Acceptance Criteria:**
Der sich drehende Kreis wird nicht mehr angezeigt.

daniel-va:
Resolves #386.
Resolves #370.

### DIFF
--- a/libs/asset-editor/src/lib/components/asset-editor-tab-page/asset-editor-tab-page.component.ts
+++ b/libs/asset-editor/src/lib/components/asset-editor-tab-page/asset-editor-tab-page.component.ts
@@ -175,7 +175,9 @@ export class AssetEditorTabPageComponent implements AfterViewInit {
 
   ngAfterViewInit(): void {
     this.referenceDataVM$.pipe(take(1)).subscribe(() => {
-      this.isFormReady = true;
+      setTimeout(() => {
+        this.isFormReady = true;
+      });
     });
   }
 
@@ -274,7 +276,7 @@ export class AssetEditorTabPageComponent implements AfterViewInit {
   }
 
   public close(): void {
-    this.routingService.navigateBack(['/']);
+    this.routingService.navigateBack(['/']).then();
   }
 
   public canLeave(): Observable<boolean> {
--- a/libs/asset-editor/src/lib/state/asset-editor.effects.ts
+++ b/libs/asset-editor/src/lib/state/asset-editor.effects.ts
@@ -1,6 +1,13 @@
 import { inject, Injectable } from '@angular/core';
 import { Router } from '@angular/router';
-import { Alert, AlertType, appSharedStateActions, filterNavigateToComponent, showAlert } from '@asset-sg/client-shared';
+import {
+  Alert,
+  AlertType,
+  appSharedStateActions,
+  filterNavigateToComponent,
+  RoutingService,
+  showAlert,
+} from '@asset-sg/client-shared';
 import { DT, isNotNull, ORD, partitionEither } from '@asset-sg/core';
 import { GeomFromGeomText } from '@asset-sg/shared';
 import * as RD from '@devexperts/remote-data-ts';
@@ -20,9 +27,10 @@ import * as actions from './asset-editor.actions';
 @UntilDestroy()
 @Injectable()
 export class AssetEditorEffects {
-  private _actions$ = inject(Actions);
-  private _assetEditorService = inject(AssetEditorService);
-  private _router = inject(Router);
+  private readonly _actions$ = inject(Actions);
+  private readonly _assetEditorService = inject(AssetEditorService);
+  private readonly _router = inject(Router);
+  private readonly routingService = inject(RoutingService);
   private readonly translateService = inject(TranslateService);
 
   validatedQueryParams = partitionEither(
@@ -109,13 +117,15 @@ export class AssetEditorEffects {
     )
   );
 
-  redirectToViewerAfterDeletion = createEffect(
-    () =>
-      this._actions$.pipe(
-        ofType(actions.handleSuccessfulDeletion),
-        switchMap(() => this._router.navigate(['']))
-      ),
-    { dispatch: false }
+  redirectToViewerAfterDeletion = createEffect(() =>
+    this._actions$.pipe(
+      ofType(actions.handleSuccessfulDeletion),
+      switchMap(async ({ assetId }) => {
+        await this.routingService.navigateToRoot();
+        return assetId;
+      }),
+      map((assetId) => appSharedStateActions.removeAssetFromSearch({ assetId }))
+    )
   );
 
   updateSearchAfterAssetChanged$ = createEffect(() =>
--- a/libs/asset-viewer/src/lib/services/viewer-params.service.ts
+++ b/libs/asset-viewer/src/lib/services/viewer-params.service.ts
@@ -41,6 +41,7 @@ export class ViewerParamsService {
 
   private async initialize(): Promise<void> {
     const paramsFromUrl = this.parseParamsFromUrl();
+
     const paramsFromStore = await firstValueFrom(this.paramsFromStore$);
 
     // We only use the values from the store if:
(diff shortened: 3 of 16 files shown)


libs/asset-editor/src/lib/state/asset-editor.effects.ts
```diff
@@ -107,13 +115,19 @@ export class AssetEditorEffects {
...
-        ofType(actions.handleSuccessfulDeletion),
-        switchMap(() => this._router.navigate(['']))
-      ),
-    { dispatch: false }
+  redirectToViewerAfterDeletion = createEffect(() =>
+    this._actions$.pipe(
+      ofType(actions.handleSuccessfulDeletion),
+      switchMap(async ({ assetId }) => {
+        await this.routingService.navigateBack(['/']);
+        await this._router.navigate([], {
```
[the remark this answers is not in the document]
daniel-va: So, the problem here is that we need to use the browser's `history.back()` to go back to the previous page after deletion, but we don't have the ability to set unset the `assetId` when going back like this.

I now added `replaceUrl` to the second navigation, which should keep the history intact. However, I'm not sure what the expected behavior would be when going back after deletion? It is not possible to go back to the asset's form, as that asset has been deleted.

libs/asset-viewer/src/lib/services/viewer-params.service.ts
```diff
@@ -0,0 +1,239 @@
...
+    );
+
+    // Ensure that the params stay up-to-date when something navigates to another page.
+    // This is mainly here so that navigations to and from the favorites viewer work.
+    this.subscription.add(
+      this.router.events.subscribe(async (event) => {
+        if (this.isUpdatingUrl || !(event instanceof NavigationEnd)) {
+          return;
+        }
+        this.params = await firstValueFrom(this.paramsFromStore$);
```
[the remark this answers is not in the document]
daniel-va: I now only update these params when the favorites actually change.

libs/asset-editor/src/lib/state/asset-editor.effects.ts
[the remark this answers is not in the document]
TIL-EBP: Why do we need to go back at all? Can't we just redirect to "/lang" and then let the new service do the rest? by then, the selectedAsset should be reomvoe from the store, so it will not be added back to the url. Or am i missing something?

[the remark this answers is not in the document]
daniel-va: Since #332, we have the requirement to go back to the previous page when closing the asset edit page. To me, it feels only natural that we use that same behavior when actually deleting the asset, as that too is kind of leaving the page.

Do you think it would make more sense to just navigate to the base path?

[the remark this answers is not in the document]
TIL-EBP: I agree that the they should be treated the same. 
BUt what i still do not understand is why we have to move back AND then navigate again?

libs/asset-viewer/src/lib/services/viewer-params.service.ts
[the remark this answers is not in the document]
TIL-EBP: I can now navigate back which updates the url correclty, but the url is not reflected in the search results (when navigating back or forward, the results stay the same)

libs/asset-editor/src/lib/components/asset-editor-tab-page/asset-editor-tab-page.component.ts
```diff
@@ -358,7 +358,7 @@ export class AssetEditorTabPageComponent {
   }
 
   public close(): void {
-    this.routingService.navigateBack(['/']);
+    this.routingService.navigateBack(['/']).then();
```
TIL-EBP, line 361: This is not a bug you introduced but: 
When i am on the favorites page and open an asset detail and then close it again, i end up on the main page, not favorites.

libs/asset-viewer/src/lib/state/asset-search/asset-search.effects.ts
TIL-EBP, line 1: NOt quite sure where to put this comment:

When trying to open the detail of an asset from the favorites view, i end up in the nor found page, because the `/favorites` does not get removed

TIL-EBP (changes requested):
The backnavigation should trigger searches again
I need to be able to open asset details from the favorite page

libs/asset-viewer/src/lib/services/viewer-params.service.ts
```diff
@@ -0,0 +1,258 @@
...
+  private subscription = new Subscription();
+
+  start(): void {
+    this.initialize().then();
+  }
+
+  stop(): void {
+    this.params = undefined as unknown as ViewerParams;
+    this.subscription.unsubscribe();
+    this.subscription = new Subscription();
```
TIL-EBP, line 39: why override the subscription?

libs/client-shared/src/lib/services/routing.service.ts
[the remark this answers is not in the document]
TIL-EBP: Edit: I can no longer reproduce this

libs/asset-editor/src/lib/components/asset-editor-tab-page/asset-editor-tab-page.component.ts
```diff
@@ -274,7 +276,7 @@ export class AssetEditorTabPageComponent implements AfterViewInit {
...
```
TIL-EBP, line 279: Closing an asset leads to weird behavior. I get redirected to a previously selected asset and only after the second clsoe am i directed to the root path

Edit: I can no longer reproduce this behavior either. Maybe my brain tricked me or i had found a super weird edge case i can nolonger reproduce. 

HOwever, when trying to navigatin back from the edit page, i am stuck there

TIL-EBP: Also, when opening an asset detail, i get an error in the console: 

Invalid lang in URL

TIL-EBP (changes requested):
It seems to be working fine now. I might have had some hallucinations because i cannot reprocduce any of the weird behaviors i believe i found. 

I will check again tomorrow and then give the approval if i still cant break it. 

Regarding the page: Before, i could navigate back to the main page, so this should be fixed anyways

libs/asset-viewer/src/lib/services/viewer-params.service.ts
daniel-va: I was assuming that the subscription could not be reused. Apparently, you can, so I have removed this now.

daniel-va: Update: this in fact needed, as I initially assumed. When reusing the unsubscribed subscription, it seems to directly unsubscribe any subscriptions added to it.

libs/asset-editor/src/lib/components/asset-editor-tab-page/asset-editor-tab-page.component.ts
daniel-va: Fixed the error in the console + the close behavior on the edit page.

PR #413, opened 2025-02-11
Fix: Cypress Tests

daniel-va:


e2e/cypress/e2e/common/login.ts
```diff
@@ -1,13 +1,18 @@
-import { assetsBaseUrl } from '../../support/config/config';
 import { Given, Then } from '@badeball/cypress-cucumber-preprocessor';
 
 Given('The user is logged in', () => {
   const userName: string = Cypress.env('basicAuthUsername') as string;
   const password: string = Cypress.env('basicAuthPassword') as string;
+  console.log('foobar', Cypress.config('baseUrl')); // Debug baseUrl
```
TIL-EBP, line 6: is this log needed still?

e2e/cypress/e2e/common/give-admin-perm.ts
```diff
@@ -1,18 +1,17 @@
 import { Given } from '@badeball/cypress-cucumber-preprocessor';
-import { assetsBaseUrl } from '../../support/config/config';
 
 Given('User has admin permissions', () => {
   const command = `docker exec -i swissgeol-assets-postgres psql -U postgres -d postgres -a -f ./give_admin_perm.sql`;
   cy.exec(command).then((result) => {
     expect(result.code).to.eq(0); // Asserts that the command ran successfully
   });
-  cy.visit(assetsBaseUrl);
+  cy.visit('/de');
```
TIL-EBP, line 8: Whats the reason for this change?

TIL-EBP (approved):

e2e/cypress/e2e/common/give-admin-perm.ts
daniel-va: When passing `localhost` directly to `cy.visit`, Cypress translates it to `127.0.0.1`. However, the development server does only serve over `localhost`, so its not possible for the test to pass locally. 

Note that now we're using the `baseUrl` config variable, which is probably the better way to do this anyways.

PR #417, opened 2025-02-13
Feature/assets 388 umbau user UI detail

### LINKED ISSUE #388: Umbau UI User Detail
**Beschreibung**

Die Detailansicht Benutzer soll entsprechend Figma erneuert werden.

https://www.figma.com/design/cEiOoOazAQZqpRY92ZhBeO/swissgeol?node-id=10301-75022&t=HhY5ioNAMv2NV5vZ-1

TIL-EBP:
resolves #388

### DIFF
--- a/apps/client-asset-sg/src/app/app.component.ts
+++ b/apps/client-asset-sg/src/app/app.component.ts
@@ -74,8 +74,7 @@ export class AppComponent {
           }
           const path = segments.slice(1).join('/');
           return !(path === 'admin' || path.startsWith(`admin/`));
-        }),
-        startWith(true)
+        })
       )
       .subscribe((showMenuBar) => {
         this.shouldShowMenuBar = showMenuBar;
--- a/apps/server-asset-sg/src/features/workgroups/workgroup.repo.ts
+++ b/apps/server-asset-sg/src/features/workgroups/workgroup.repo.ts
@@ -144,5 +144,6 @@ const parse = (data: SelectedWorkgroup): Workgroup => {
     name: data.name,
     users,
     disabledAt: data.disabled_at,
+    numberOfAssets: data.assets.length,
   };
 };
--- a/libs/admin/src/lib/components/add-workgroup-to-user-dialog/add-workgroup-to-user-dialog.component.ts
+++ b/libs/admin/src/lib/components/add-workgroup-to-user-dialog/add-workgroup-to-user-dialog.component.ts
@@ -1,0 +1,68 @@
+import { Component, inject } from '@angular/core';
+import { MAT_DIALOG_DATA, MatDialogRef } from '@angular/material/dialog';
+import { Filter } from '@asset-sg/client-shared';
+import { User, Workgroup, WorkgroupId } from '@asset-sg/shared/v2';
+import { Store } from '@ngrx/store';
+import { Role } from '@prisma/client';
+import * as actions from '../../state/admin.actions';
+import { AppStateWithAdmin } from '../../state/admin.reducer';
+
+@Component({
+  selector: 'asset-sg-add-workgroup-to-user-dialog',
+  templateUrl: './add-workgroup-to-user-dialog.component.html',
+  styleUrls: ['./add-workgroup-to-user-dialog.component.scss'],
+  standalone: false,
+})
+export class AddWorkgroupToUserDialogComponent {
+  public user?: User;
+  public workgroups: Filter<WorkgroupId>[] = [];
+  public roleSelectors: Filter<Role>[] = [];
+  public shouldShowError = false;
+  private readonly store = inject(Store<AppStateWithAdmin>);
+  private readonly dialogRef = inject(MatDialogRef<AddWorkgroupToUserDialogComponent>);
+  private readonly data = inject<{ workgroups: Workgroup[]; user: User | null }>(MAT_DIALOG_DATA);
+  private selectedWorkgroupIds: WorkgroupId[] = [];
+  private selectedRole: Role = Role.Viewer;
+
+  constructor() {
+    this.user = this.data.user ?? undefined;
+    this.workgroups = this.data.workgroups
+      .filter((workgroup) => !this.user?.roles.has(workgroup.id))
+      .map((workgroup) => ({
+        displayValue: workgroup.name,
+        value: workgroup.id,
+      }));
+    this.roleSelectors = Object.values(Role).map((role) => ({ displayValue: role, value: role }));
+  }
+
+  public setSelectedRole(role: Filter<Role>[]) {
+    this.selectedRole = role[0].value;
+  }
+
+  public setSelectedWorkgroups(selectedWorkgroups: Filter<WorkgroupId>[]) {
+    this.selectedWorkgroupIds = selectedWorkgroups.map((workgroup) => workgroup.value);
+  }
+
+  public close() {
+    this.dialogRef.close();
+  }
+
+  public updateWorkgroupsOfUser() {
+    if (this.selectedWorkgroupIds.length === 0) {
+      this.shouldShowError = true;
+      return;
+    }
+    if (this.user == null) {
+      return;
+    }
+    const roles = this.user.roles;
+    for (const workgroupId of this.selectedWorkgroupIds) {
+      roles.set(workgroupId, this.selectedRole);
+    }
+    this.store.dispatch(actions.updateUser({ user: { ...this.user, roles } }));
+    this.shouldShowError = false;
+    this.dialogRef.close();
+  }
+
+  protected readonly Role = Role;
+}
--- a/libs/admin/src/lib/components/add-workgroup-user-dialog/add-workgroup-user-dialog.component.ts
+++ b/libs/admin/src/lib/components/add-workgroup-user-dialog/add-workgroup-user-dialog.component.ts
@@ -72,7 +72,7 @@ export class AddWorkgroupUserDialogComponent implements OnInit {
     if (this.mode === 'edit') {
       this.store.dispatch(actions.updateWorkgroup({ workgroupId: this.workgroup.id, workgroup }));
     } else {
-      this.store.dispatch(actions.setWorkgroup({ workgroup: { ...workgroup, id: -1 } }));
+      this.store.dispatch(actions.setWorkgroup({ workgroup: { ...workgroup, id: -1, numberOfAssets: 0 } }));
     }
     this.dialogRef.close();
   }
--- a/libs/admin/src/lib/components/admin-page/admin-page.component.html
+++ b/libs/admin/src/lib/components/admin-page/admin-page.component.html
@@ -4,11 +4,12 @@
     <svg-icon key="arrow-left" class="close" />
   </button>
   @if (isWorkgroupsPage) {
-  <h3>{{ (selectedWorkgroup$ | async)?.name }}</h3>
+  <h2>{{ (selectedWorkgroup$ | async)?.name }}</h2>
   } @else if (isUsersPage) {
-  <h3>{{ (selectedUser$ | async)?.email }}</h3>
+  <h2>{{ (selectedUser$ | async)?.email }}</h2>
+  <asset-sg-chip type="secondary"><span translate>admin.user</span></asset-sg-chip>
   } @else {
-  <h3>{{ "menuBar.settings" | translate }}</h3>
+  <h2 translate="">menuBar.settings</h2>
   }
 </div>
 <div class="main">
--- a/libs/admin/src/lib/components/admin-page/admin-page.component.scss
+++ b/libs/admin/src/lib/components/admin-page/admin-page.component.scss
@@ -2,7 +2,7 @@
 
 $sub-navigation-bar-height: 84px;
 
-::ng-deep .navigation .navigation__link {
+:host ::ng-deep .navigation .navigation__link {
   .mdc-tab__text-label {
     color: variables.$cyan-09;
   }
@@ -36,7 +36,7 @@ $sub-navigation-bar-height: 84px;
   display: flex;
   align-items: center;
   justify-content: flex-start;
-  gap: 12px;
+  gap: 18px;
   padding-left: 24px;
 
   .back-button {
@@ -44,6 +44,12 @@ $sub-navigation-bar-height: 84px;
     width: 36px;
     padding: 0;
   }
+}
+
+h2 {
+  margin: 0;
+  font-size: 26px;
+  line-height: 28px;
 }
 
 .main {
--- a/libs/admin/src/lib/components/detail-section/detail-section.component.html
+++ b/libs/admin/src/lib/components/detail-section/detail-section.component.html
@@ -1,0 +1,11 @@
+<div class="container">
+  <div class="header">
+    <h2>
+      {{ title }}
+    </h2>
+    <ng-content select="[actions]"></ng-content>
+  </div>
+  <div class="content">
+    <ng-content select="[content]"></ng-content>
+  </div>
+</div>
--- a/libs/admin/src/lib/components/user-edit/user-edit.component.html
+++ b/libs/admin/src/lib/components/user-edit/user-edit.component.html
@@ -1,101 +1,102 @@
-<mat-card class="card">
-  <div class="user-edit">
-    <form [formGroup]="formGroup" class="form">
-      <div *ngIf="user" class="header">
-        <h2 class="email">
-          {{ user.email }}
-        </h2>
-      </div>
-      <div class="user-data">
-        <div class="user-data__group">
-          <mat-form-field class="lang-form-field" subscriptSizing="dynamic">
-            <mat-label translate>admin.userPage.lang</mat-label>
-            <mat-select formControlName="lang">
-              <mat-option value="de" translate><span translate>admin.languages.de</span></mat-option>
-              <mat-option value="en" translate><span translate>admin.languages.en</span></mat-option>
-              <mat-option value="fr" translate><span translate>admin.languages.fr</span></mat-option>
-              <mat-option value="it" translate><span translate>admin.languages.it</span></mat-option>
-            </mat-select>
-          </mat-form-field>
-        </div>
-        <div class="user-data__group">
-          <mat-checkbox formControlName="isAdmin" color="primary">
-            <span translate>admin.userPage.admin</span>
-          </mat-checkbox>
-        </div>
-      </div>
-      <div class="workgroups">
-        <div class="workgroups__header">
-          <h3 translate>admin.workgroups</h3>
-          <form>
-            <mat-form-field>
-              <mat-label translate>admin.userPage.addWorkgroups</mat-label>
-              <input
-                type="text"
-                [placeholder]="'admin.name' | translate"
-                matInput
-                [formControl]="workgroupAutoCompleteControl"
-                [matAutocomplete]="auto"
-              />
-              <button
-                class="input-clear-button"
-                asset-sg-icon-button
-                matSuffix
-                cdkMonitorElementFocus
-                (click)="resetWorkgroupSearch()"
-              >
-                <svg-icon key="close"></svg-icon>
-              </button>
-              <mat-autocomplete #auto="matAutocomplete">
-                <mat-option
-                  *ngFor="let workgroup of filteredWorkgroups"
-                  [value]="workgroup.name"
-                  (onSelectionChange)="addWorkgroupRole($event, workgroup.id)"
-                  [disabled]="isUserPartOfWorkgroup(workgroup.id)"
-                >
-                  {{ workgroup.name }}
-                </mat-option>
-              </mat-autocomplete>
-            </mat-form-field>
-          </form>
-        </div>
-        <div>
-          <table mat-table [dataSource]="userWorkgroups$" class="mat-elevation-z1">
-            <!-- ID Column -->
-            <ng-container matColumnDef="name">
-              <th mat-header-cell *matHeaderCellDef>Name</th>
-              <td mat-cell *matCellDef="let workgroup">{{ workgroup.name }}</td>
-            </ng-container>
-
-            <!-- Role Column -->
-            <ng-container matColumnDef="role">
-              <th mat-header-cell *matHeaderCellDef translate>admin.role</th>
-              <td mat-cell *matCellDef="let workgroup">
-                <mat-form-field appearance="fill" subscriptSizing="dynamic" class="workgroups__table__form-field">
-                  <mat-label translate>admin.role</mat-label>
-                  <mat-select [value]="workgroup.role" (selectionChange)="updateWorkgroupRole($event, workgroup.id)">
-                    <mat-option *ngFor="let role of roles" [value]="role">
-                      {{ role }}
-                    </mat-option>
-                  </mat-select>
-                </mat-form-field>
-              </td>
-            </ng-container>
-
-            <ng-container matColumnDef="actions">
-              <th mat-header-cell *matHeaderCellDef translate>admin.actions</th>
-              <td mat-cell *matCellDef="let workgroup">
-                <button (click)="removeWorkgroupRole(workgroup.id)">
-                  <svg-icon key="delete"></svg-icon>
-                </button>
-              </td>
-            </ng-container>
-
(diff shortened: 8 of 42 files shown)


libs/admin/src/lib/components/admin-page/admin-page.component.html
```diff
@@ -1,21 +1,33 @@
...
+  <button asset-sg-primary (click)="navigateBack(currentLang)" class="back-button">
+    <svg-icon key="arrow-left" class="close" />
+  </button>
+  @if (isWorkgroupsPage) {
+  <h2>{{ (selectedWorkgroup$ | async)?.name }}</h2>
+  } @else if (isUsersPage) {
+  <h2>{{ (selectedUser$ | async)?.email }}</h2>
+  <asset-sg-chip [type]="'secondary'">{{ "admin.user" | translate }}</asset-sg-chip>
+  } @else {
+  <h2>{{ "menuBar.settings" | translate }}</h2>
```
daniel-va, line 12: You could move these `translate` pipes into the elements.
Also, string attributes can be passed without `[` brackets `]`.
```suggestion
  <asset-sg-chip type="secondary" translate>admin.user</asset-sg-chip>
  } @else {
  <h2 translate>menuBar.settings</h2>
```

libs/admin/src/lib/components/admin-page/admin-page.component.scss
```diff
@@ -1,5 +1,7 @@
 @use "../../styles/variables";
 
+$sub-navigation-bar-height: 84px;
+
 ::ng-deep .navigation .navigation__link {
```
daniel-va, line 5: `::ng-deep` without any prefix is essentially just global css. This was not added in this PR, but we might just as well fix it while we're here.
```suggestion
:host ::ng-deep .navigation .navigation__link {
```

libs/admin/src/lib/components/add-workgroup-to-user-dialog/add-workgroup-to-user-dialog.component.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  public ngOnInit() {
+    this.subscription.add(
+      this.selectedWorkgroupIds$
+        .pipe(
+          combineLatestWith(this.selectedRole$),
+          tap(([workgroupIds, role]) => {
+            for (const workgroupId of workgroupIds) {
+              this.user?.roles.set(workgroupId, role);
+            }
```
daniel-va, line 49: I'm not a fan of mutating an input object directly, especially if it is managed in the store. This has the potential to be a huge headache if there is ever a need to debug this.

My suggestion: Change the `selected*` variables to be normal fields instead of observables, and update the user in the dispatch within `updateWorkgroupsOfUser`.

apps/client-asset-sg/src/app/app.component.ts
```diff
@@ -57,6 +60,30 @@ export class AppComponent {
...
         setCssCustomProperties(document.documentElement, ['font-size', fontSize]);
       });
+
+    this.router.events
+      .pipe(
+        untilDestroyed(this),
+        filter((event) => event instanceof NavigationEnd),
+        startWith(() => undefined),
+        map(() => {
+          const segments = (this.router.getCurrentNavigation() ?? this.router.lastSuccessfulNavigation)?.finalUrl?.root
```
daniel-va, line 70: I get this error on the search page:
image

libs/admin/src/lib/components/detail-section/detail-section.component.html
```diff
@@ -0,0 +1,17 @@
+<div class="container">
+  <div class="header">
+    <h2>
+      {{ title }}
+    </h2>
+    @if (hasButton) {
+    <button asset-sg-primary (click)="onButtonClick()">
+      {{ buttonTitle }}
+      <svg-icon [key]="buttonIcon"></svg-icon>
+    </button>
```
daniel-va, line 10: I would personally not have this button in here, but instead just use an `<ng-content />` with a `select`, and then have the component's call sides handle this button by themselves.

libs/admin/src/lib/components/user-edit/user-edit.component.html
```diff
@@ -1,101 +1,105 @@
...
+  (buttonClick)="openAddWorkgroupToUserDialog()"
+>
+  <asset-sg-search-and-filter
+    (searchTermChanged)="setSearchTerm($event)"
+    (shouldShowFiltersChanged)="toggleFilters($event)"
+  ></asset-sg-search-and-filter>
+  @if (shouldShowFilters) {
+  <div class="filters">
+    <asset-sg-filter-selector
+      title="{{ 'admin.role' | translate }}"
```
daniel-va, line 43: This reads better, IMO.
```suggestion
      [title]="'admin.role' | translate"
```

libs/admin/src/lib/components/user-edit/user-edit.component.ts
```diff
@@ -1,72 +1,69 @@
...
+  protected readonly COLUMNS = ['name', 'amount', 'role', 'isActive', 'actions'];
+  languageSelector: Filter[] = [
+    { displayValue: { key: 'admin.languages.de' }, value: 'de' },
+    { displayValue: { key: 'admin.languages.en' }, value: 'en' },
+    { displayValue: { key: 'admin.languages.fr' }, value: 'fr' },
+    { displayValue: { key: 'admin.languages.it' }, value: 'en' },
+  ];
 
-  protected readonly COLUMNS = ['name', 'role', 'actions'];
+  public roleSelector: Filter[] = [];
```
daniel-va, line 49:
```suggestion
  public roleSelectors: Filter[] = [];
```

libs/admin/src/lib/components/user-edit/user-edit.component.ts
```diff
@@ -75,35 +72,89 @@ export class UserEditComponent implements OnInit, OnDestroy {
...
+
+  public setFilters(selectedValues: Filter[], key: keyof User) {
+    const activeFilters = this.activeFilters$.value.set(
+      key,
+      selectedValues.map((it) => it.value)
+    );
+    this.activeFilters$.next(activeFilters);
+  }
+
+  private matchUsersBySearchTerm(userWorkgroup: UserWorkgroup, searchTerm: string): boolean {
```
daniel-va, line 109:
```suggestion
  private matchWorkgroupBySearchTerm(userWorkgroup: UserWorkgroup, searchTerm: string): boolean {
```

libs/admin/src/lib/components/user-edit/user-edit.component.ts
```diff
@@ -75,35 +72,89 @@ export class UserEditComponent implements OnInit, OnDestroy {
...
+    const searchTermLowerCase = searchTerm.toLowerCase();
+    return Object.values(userWorkgroup).some((value) => {
+      return value.toString().toLowerCase().includes(searchTermLowerCase);
+    });
+  }
+
+  private filterWorkgroupsByAttribute(role: Role, filters: Map<keyof User, (string | number | boolean)[]>): boolean {
```
daniel-va, line 116: Bad naming here - this method doesn't filter workgroups, it checks if the `role` matches `filters`.

libs/admin/src/lib/components/user-edit/user-edit.component.ts
```diff
@@ -75,35 +72,89 @@ export class UserEditComponent implements OnInit, OnDestroy {
...
+        return true;
+      }
+      return values.some((value) => {
+        return value === role;
+      });
+    });
   }
 
-  public updateWorkgroupRole(event: MatSelectChange, workgroupId: WorkgroupId) {
+  public readonly userWorkgroups$: Observable<UserWorkgroup[]> = this.user$.pipe(
```
daniel-va, line 127: Move this up to to the instance variables.

libs/admin/src/lib/components/user-edit/user-edit.component.ts
```diff
@@ -75,35 +72,89 @@ export class UserEditComponent implements OnInit, OnDestroy {
...
+    combineLatestWith(this.workgroups$),
```
daniel-va, line 128: I prefer using `combineLatest` if possible, as that makes it clear that all combined sources can cause the observable to trigger.

libs/admin/src/lib/components/user-edit/user-edit.component.ts
```diff
@@ -1,72 +1,69 @@
...
 import { Role } from '@prisma/client';
-import { asapScheduler, filter, map, Observable, startWith, Subscription, withLatestFrom } from 'rxjs';
+import { BehaviorSubject, combineLatestWith, filter, map, Observable, Subscription, tap } from 'rxjs';
 import * as actions from '../../state/admin.actions';
 import { AppStateWithAdmin } from '../../state/admin.reducer';
-import { selectSelectedUser } from '../../state/admin.selector';
+import { selectSelectedUser, selectWorkgroups } from '../../state/admin.selector';
+import { AddWorkgroupToUserDialogComponent } from '../add-workgroup-to-user-dialog/add-workgroup-to-user-dialog.component';
+
+export type UserWorkgroup = SimpleWorkgroup & { role: Role; isActive: boolean; numberOfAssets: number };
```
daniel-va, line 20: This name confuses me somewhat. I'd prefer something like `WorkgroupOfUser`. Feel free to use whatever feels best to you, just my two cents.

libs/admin/src/lib/components/user-edit/user-edit.component.ts
```diff
@@ -116,6 +167,39 @@ export class UserEditComponent implements OnInit, OnDestroy {
...
+          return this.compare(a.name, b.name, isAsc);
+        default:
+          return 0;
+      }
+    });
+  }
+
+  private compare(a: string, b: string, isAsc: boolean) {
+    return (a < b ? -1 : 1) * (isAsc ? 1 : -1);
+  }
```
daniel-va, line 189: This could be a free function instead of a method.

libs/admin/src/lib/components/user-edit/user-edit.component.ts
```diff
@@ -116,6 +167,39 @@ export class UserEditComponent implements OnInit, OnDestroy {
...
+
+  public languageChanged(updatedValue: Filter[]) {
```
daniel-va, line 191: We're currently lacking a standard naming scheme for our event handlers. Maybe we could just go the same route as in viewer, e.g. `handleLanguageChanged`?

libs/admin/src/lib/components/user-edit/user-edit.component.ts
```diff
@@ -116,6 +167,39 @@ export class UserEditComponent implements OnInit, OnDestroy {
...
+    if (typeof updatedValue[0].value === 'string' && this.user) {
+      this.updateUser({ ...this.user, lang: updatedValue[0].value });
+    }
+  }
+
+  public isAdminChanged(event: MatCheckboxChange) {
```
daniel-va, line 197: This is a weird name for an event handler, it should probably not start with `is`.

libs/admin/src/lib/components/user-edit/user-edit.component.ts
```diff
@@ -131,67 +215,49 @@ export class UserEditComponent implements OnInit, OnDestroy {
...
-      { emitEvent: false }
-    );
-  }
-
   private initSubscriptions() {
     this.subscriptions.add(
       this.user$.subscribe((user) => {
         this.user = user;
-        this.initializeForm();
+        console.log('user', user);
```
daniel-va, line 222:
```suggestion
```

libs/admin/src/lib/components/user-edit/user-edit.component.ts
```diff
@@ -131,67 +215,49 @@ export class UserEditComponent implements OnInit, OnDestroy {
...
-      this.isCurrentUser$.subscribe((isCurrentUser) => {
-        if (isCurrentUser) {
-          this.formGroup.controls.isAdmin.disable();
-        } else {
-          this.formGroup.controls.isAdmin.enable();
-        }
-      })
+      this.searchTerm$
+        .pipe(
+          combineLatestWith(this.activeFilters$),
```
daniel-va, line 250: Here, too, I would use `combineLatest` instead of `combineLatestWith`.

libs/admin/src/lib/components/users/users.component.html
```diff
@@ -1,49 +1,86 @@
...
+          mat-cell
+          *matCellDef="let user"
+          [matTooltip]="formatWorkgroupsTooltip(user.roles)"
+          matTooltipClass="workgroups-tooltip"
+          matTooltipPosition="left"
+        >
+          <mat-chip-set>
+            @for (workgroup of getWorkgroupsOfUser(user); track workgroup.id) {
+            <asset-sg-chip>{{ workgroup.name }}</asset-sg-chip>
+            } @if (user.roles.size > WORKGROUP_DISPLAY_COUNT) {
```
daniel-va, line 58:
```suggestion
            }
            @if (user.roles.size > WORKGROUP_DISPLAY_COUNT) {
```

libs/admin/src/lib/components/users/users.component.scss
```diff
@@ -8,20 +8,30 @@
...
-  margin-bottom: 1rem;
+  gap: 12px;
 }
 
-.table__header {
-  background-color: variables.$grey-03;
+.mat-column-firstName,
+.mat-column-lastName,
+.mat-column-email {
+  width: 14%;
```
daniel-va, line 21: Sizing width by percentage is always a bit of a pitfall. Make sure this works as expected when scaling down the window. I would recommended fixed widths, or flex/grid if possible.

libs/admin/src/lib/components/users/users.component.ts
```diff
@@ -33,27 +55,89 @@ export class UsersComponent implements OnInit, OnDestroy {
...
+
+  public setFilters(selectedValues: Filter[], key: keyof User) {
+    const activeFilters = this.activeFilters$.value.set(
+      key,
+      selectedValues.map((it) => it.value)
+    );
+    this.activeFilters$.next(activeFilters);
+  }
+
+  private matchUsersBySearchTerm(user: User, searchTerm: string): boolean {
```
daniel-va, line 89:
```suggestion
  private matchUserBySearchTerm(user: User, searchTerm: string): boolean {
```

libs/admin/src/lib/components/users/users.component.ts
```diff
@@ -33,27 +55,89 @@ export class UsersComponent implements OnInit, OnDestroy {
...
       }
-      const [workgroupId, role] = value;
-      const workgroup = this.workgroups.get(workgroupId);
-      if (workgroup == null) {
-        continue;
+      return value.toString().toLowerCase().includes(searchTerm);
+    });
+  }
+
+  private filterUsersByAttribute(user: User, filters: Map<keyof User, (string | number | boolean)[]>): boolean {
```
daniel-va, line 101: Bad name, again, as the method doesn't filter.

libs/client-shared/src/lib/components/filter-selector/filter-selector.component.ts
```diff
@@ -0,0 +1,52 @@
...
+    FormsModule,
+    MatFormField,
+    SmartTranslatePipe,
+    MatHint,
+  ],
+})
+export class FilterSelectorComponent implements OnInit {
+  @Input() public values: Filter[] = [];
+  @Input() public title = '';
+  @Input() public multiple = true;
```
daniel-va, line 34: The default for a boolean field should always be `false`, as that makes it more natural to enable it. Also, take a look at `coerceBooleanProperty` (e.g. in the `MenuBarItemComponent`), which would make it possible to write this input like this:
```html
<asset-sg-filter-selector multiple>
```

libs/shared/v2/src/lib/schemas/workgroup.schema.ts
```diff
@@ -19,6 +19,9 @@ export class WorkgroupDataSchema extends Schema implements WorkgroupData {
   @IsNullable()
   @Type(() => Date)
   disabledAt!: Date | null;
+
+  @IsArray()
+  assets!: number[];
```
daniel-va, line 24:
```suggestion
  @IsArray()
  @Type(() => Number)
  assets!: number[];
```

daniel-va (changes requested):

apps/server-asset-sg/src/features/workgroups/workgroup.repo.ts
```diff
@@ -144,5 +144,6 @@ const parse = (data: SelectedWorkgroup): Workgroup => {
     name: data.name,
     users,
     disabledAt: data.disabled_at,
+    assets: data.assets.map((asset) => asset.assetId),
```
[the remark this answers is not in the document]
TIL-EBP: I do need some way of accessing the number of assets. The alternative would be to add another parameter like we did for simplified, so this datta only gets sent for workgroups on the admin page and not for workgroups in the filters

libs/admin/src/lib/components/add-workgroup-to-user-dialog/add-workgroup-to-user-dialog.component.ts
```diff
@@ -0,0 +1,86 @@
...
+        )
+        .subscribe()
+    );
+  }
+
+  public ngOnDestroy() {
+    this.subscription.unsubscribe();
+  }
+
+  public setSelectedRole(role: Filter[]) {
```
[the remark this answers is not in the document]
TIL-EBP: So would you have two different emiiters then? because oterwise, i would have it emit either Arrays or single values, meaning the listener would have to check everytime whether they received an array or not

libs/admin/src/lib/components/add-workgroup-to-user-dialog/add-workgroup-to-user-dialog.component.ts
```diff
@@ -0,0 +1,86 @@
...
+    this.selectedRole$.next(role[0].value as Role);
+  }
+
+  public setSelectedWorkgroups(selectedWorkgroups: Filter[]) {
+    this.selectedWorkgroupIds$.next(selectedWorkgroups.map((workgroup) => workgroup.value as WorkgroupId));
+  }
+
+  public close() {
```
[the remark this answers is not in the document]
TIL-EBP: why?

libs/admin/src/lib/components/admin-page/admin-page.component.html
TIL-EBP: Adding the translate pipe directly to the `asset-sg-chip` component does not work. I need to add an additional span, which then does not really shorten anything nor make it more readable IMO

libs/admin/src/lib/components/add-workgroup-to-user-dialog/add-workgroup-to-user-dialog.component.ts
TIL-EBP: So do you want to create a completely new map and never set any values to the user.roles map?

libs/admin/src/lib/components/user-edit/user-edit.component.ts
TIL-EBP: `combineLatest` is deprecated

libs/admin/src/lib/components/user-edit/user-edit.component.ts
TIL-EBP: The naming comes from the attribute which is called `isAdmin` so according to your other comment, I will rename it to `handleIsAdminChanged`

libs/admin/src/lib/components/user-edit/user-edit.component.ts
TIL-EBP: See comment above

apps/client-asset-sg/src/app/app.component.ts
```diff
@@ -57,6 +60,30 @@ export class AppComponent {
...
+            .children['primary'].segments;
+          if (segments == null || segments.length === 1) {
+            return true;
+          }
+          const path = segments.slice(1).join('/');
+          return !(path === 'admin' || path.startsWith(`admin/`));
+        }),
+        startWith(true)
```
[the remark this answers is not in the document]
TIL-EBP: now. the first one can be omitted. 

Regarding your other comment, i am not quite sure what you mean by that

libs/admin/src/lib/components/add-workgroup-user-dialog/add-workgroup-user-dialog.component.ts
```diff
@@ -67,6 +67,7 @@ export class AddWorkgroupUserDialogComponent implements OnInit {
     const workgroup: WorkgroupData = {
       name: this.workgroup.name,
       disabledAt: this.workgroup.disabledAt,
+      assets: this.workgroup.assets,
```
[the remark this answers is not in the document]
TIL-EBP: I see. So  I could move it to the Workgroup Schema instead. BUt maybe at some point we would like to be able to edit the Assets in a workgroup directly in the workgroup edit ui, and then we woul dhave to move it back to the WOrkgroupDataSchema

libs/admin/src/lib/components/add-workgroup-to-user-dialog/add-workgroup-to-user-dialog.component.ts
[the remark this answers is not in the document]
TIL-EBP: Nevermind, this component no longer needs the lifecycle hooks anyways

libs/admin/src/lib/components/admin-page/admin-page.component.html
```diff
@@ -7,9 +7,9 @@
   <h2>{{ (selectedWorkgroup$ | async)?.name }}</h2>
   } @else if (isUsersPage) {
   <h2>{{ (selectedUser$ | async)?.email }}</h2>
-  <asset-sg-chip [type]="'secondary'">{{ "admin.user" | translate }}</asset-sg-chip>
+  <asset-sg-chip type="secondary"><span translate>admin.user</span></asset-sg-chip>
```
daniel-va, line 10: Yeah okay, that is not really an improvement. I assumed the translate would work...

libs/admin/src/lib/components/detail-section/detail-section.component.html
```diff
@@ -3,15 +3,9 @@
...
       {{ title }}
     </h2>
-    @if (hasButton) {
-    <button asset-sg-primary (click)="onButtonClick()">
-      {{ buttonTitle }}
-      <svg-icon [key]="buttonIcon"></svg-icon>
-    </button>
-
-    }
+    <ng-content select="[button]"></ng-content>
```
daniel-va, line 6: I would call this select `prefix` or `actions`, as it doesn't necessarily contain a button.
```suggestion
    <ng-content select="[prefix]"></ng-content>
```

daniel-va (approved):
# squash

apps/client-asset-sg/src/app/app.component.ts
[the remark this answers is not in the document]
daniel-va: I just meant that in a lot places, we parse the URL by hand (again and again), instead of having utility functions or even a service for it. I would not change this in this PR, but would like if we would take up a CQ issue for that at some point.

apps/server-asset-sg/src/features/workgroups/workgroup.repo.ts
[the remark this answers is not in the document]
daniel-va: If you just need the number, then why don't you just send that instead of the entire list of asset ids?

libs/admin/src/lib/components/add-workgroup-to-user-dialog/add-workgroup-to-user-dialog.component.ts
daniel-va: Yes, your change is what I meant. The way it is now, it is much clearer to understand what is going on, and we don't run the risk of mutating something that some other place might expect to never change.

apps/server-asset-sg/src/features/workgroups/workgroup.repo.ts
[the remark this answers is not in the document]
TIL-EBP: just in case we need to be able to edit assetss from within a workgroup in the future. BUt yes, just sending the number of assets would be sufficient in this case

PR #418, opened 2025-02-17
update the first and last name of a user correctly

TIL-EBP:
adds the first and lastname of a user to the database if it already exists


daniel-va (approved):
funny stuff

PR #427, opened 2025-02-19
Feature/assets 390 umbau UI workgroups uebersicht

### LINKED ISSUE #390: Umbau UI Arbeitsgruppen Übersicht
**Beschreibung**

Die Übersicht von Arbeitsgruppen soll entsprechend Figma erneuert werden.

https://www.figma.com/design/cEiOoOazAQZqpRY92ZhBeO/swissgeol?node-id=10301-40865&t=HhY5ioNAMv2NV5vZ-1

TIL-EBP:
resolves #390


libs/admin/src/lib/components/admin-page/admin-page.component.ts
```diff
@@ -31,17 +31,21 @@ export class AdminPageComponent implements OnInit {
...
   }
 
-  public get isWorkgroupsPage(): boolean {
+  public get isWorkgroupDetailPage(): boolean {
     return this.router.url.includes('/workgroups/');
   }
 
+  public get isWorkgroupPage(): boolean {
+    return this.router.url.endsWith('/workgroups');
+  }
```
daniel-va, line 47: This is the same method with a different name.

libs/admin/src/lib/components/users/users.component.ts
```diff
@@ -214,6 +214,6 @@ export class UsersComponent implements OnInit, OnDestroy, AfterViewInit {
   }
 }
 
-export function compare(a: string, b: string, isAsc: boolean) {
+export function compare<T>(a: T, b: T, isAsc: boolean) {
```
daniel-va, line 217: Might be a good idea to limit this type to `number | string`, just so we don't confuse this method as being able to compare anything.

libs/admin/src/lib/components/workgroups/workgroups.component.ts
```diff
@@ -1,26 +1,177 @@
...
+        });
+      }
+      if (key === 'disabledAt') {
+        return false;
+      }
+      return value.toString().toLowerCase().includes(searchTerm);
+    });
+  }
+
+  getNUmberOfUsersPerRoleForWorkgroup(workgroup: Workgroup): UsersPerRole {
```
daniel-va, line 109:
```suggestion
  getNumberOfUsersPerRoleForWorkgroup(workgroup: Workgroup): UsersPerRole {
```

libs/admin/src/lib/state/admin.effects.ts
```diff
@@ -72,6 +72,15 @@ export class AdminEffects {
     )
   );
 
+  public deleteWorkgroup$ = createEffect(() =>
+    this.actions$.pipe(
+      ofType(actions.deleteWorkgroup),
+      switchMap(({ workgroupId }) =>
+        this.adminService.deleteWorkgroup(workgroupId).pipe(map(() => actions.listWorkgroups()))
```
daniel-va, line 79: Can't we just remove the workgroup from the list instead of fetching all workgroups again?

libs/client-shared/src/lib/components/button/button.component.ts
```diff
@@ -46,7 +46,7 @@ export class ButtonComponent implements AfterViewInit, OnDestroy {
   private _classes: string | undefined;
 
   private _disabled: true | undefined;
-  @HostBinding('[attr.disabled]')
+  @HostBinding('attr.disabled')
   @HostBinding('[attr.aria-disabled]')
```
daniel-va, line 50: Either remove the brackets to make this binding work, or remove it altogether.

libs/client-shared/src/lib/components/chip/chip.component.scss
```diff
@@ -13,6 +13,12 @@
   }
 }
 
+:host(.disabled) {
+  .mat-mdc-standard-chip.mdc-evolution-chip--disabled {
+    background-color: #acb4bd;
+  }
+}
```
daniel-va, line 20:
```suggestion
:host(.disabled) .mat-mdc-standard-chip.mdc-evolution-chip--disabled {
  background-color: #acb4bd;
}
```

libs/admin/src/lib/components/workgroups/workgroups.component.html
```diff
@@ -1,36 +1,80 @@
...
+    </ng-container>
+    <ng-container matColumnDef="actions">
+      <th mat-header-cell *matHeaderCellDef translate></th>
+      <td mat-cell *matCellDef="let workgroup">
+        <button
+          [disabled]="workgroup.numberOfAssets > 0"
+          asset-sg-secondary
+          isIcon
+          (click)="deleteWorkgroup($event, workgroup.id)"
+          [matTooltip]="
```
daniel-va, line 60: I would position that tooltip to the left - it blocks the line above the hovered one in its current state.

daniel-va (changes requested):
Some small things that should be looked at, but overall looking good 😄 
+ it would probably be a good idea to extract some of the common behavior of the users and workgroups table into a new component.

libs/admin/src/lib/components/admin-page/admin-page.component.ts
TIL-EBP: NOt quite, one is checking if it is a WorkgroupDetailPage (does anything follow after the` /workgroups`, the other checks if it is the workgroups overview page (url ends after  `/workgroups`

libs/admin/src/lib/state/admin.effects.ts
TIL-EBP: Done

TIL-EBP:
Regarding the extraction of common behavior, I will do this in the next PR for the workgroup detail

libs/admin/src/lib/state/admin.effects.ts
```diff
@@ -76,7 +76,9 @@ export class AdminEffects {
     this.actions$.pipe(
       ofType(actions.deleteWorkgroup),
       switchMap(({ workgroupId }) =>
-        this.adminService.deleteWorkgroup(workgroupId).pipe(map(() => actions.listWorkgroups()))
+        this.adminService
+          .deleteWorkgroup(workgroupId)
+          .pipe(map(() => actions.removeWorkgroupAfterDelete({ workgroupId })))
       )
```
daniel-va, line 82: For readability, I pretty much always prefer to move `pipe` out of `switchMap`:
```suggestion
      switchMap(({ workgroupId }) => this.adminService.deleteWorkgroup(workgroupId)),
      map(() => actions.removeWorkgroupAfterDelete({ workgroupId })),
```

daniel-va (approved):

libs/admin/src/lib/state/admin.effects.ts
TIL-EBP: Outside the switchamp i no longer have access to the `workgroupId`. I would have to return it fro within the adminService, which seems unnecessary to me

daniel-va: very true, did not see that

PR #428, opened 2025-02-24
Feature/assets 391 umbau UI workgroups detail

### LINKED ISSUE #391: Umbau UI Arbeitsgruppen Detail
**Beschreibung**

Die Detailansicht von Arbeitsgruppen soll entsprechend Figma erneuert werden.

https://www.figma.com/design/cEiOoOazAQZqpRY92ZhBeO/swissgeol?node-id=10301-86791&t=HhY5ioNAMv2NV5vZ-1

TIL-EBP:
resolves #391


libs/client-shared/src/lib/components/text-input/text-input.component.ts
```diff
@@ -0,0 +1,22 @@
...
+import { TranslateModule } from '@ngx-translate/core';
+
+@Component({
+  selector: 'asset-sg-text-input',
+  templateUrl: './text-input.component.html',
+  styleUrls: ['./text-input.component.scss'],
+  standalone: true,
+  imports: [SvgIconComponent, TranslateModule, FormsModule],
+})
+export class TextInputComponent {
```
daniel-va, line 13: It's nice that we get to implementing our own ui components now. I would like if, for input elements, we could write them to integrate with stuff like `ngModel` and reactive forms.

Have a look at [ControlValueAccessor](https://angular.dev/api/forms/ControlValueAccessor). It serves as the basis for such behavior in Angular.

libs/client-shared/src/lib/components/text-input/text-input.component.ts
```diff
@@ -0,0 +1,22 @@
...
+  @Input() public term = '';
+  @Input() public hasSearchIcon = false;
```
daniel-va, line 15: I would being able to pass any icon I want here instead of being restricted to the search icon.

libs/client-shared/src/lib/components/text-input/text-input.component.scss
```diff
@@ -14,6 +14,10 @@
   }
 }
 
+.hasSearchIcon {
```
daniel-va, line 17: Class names should be `kebap-case`.

libs/client-shared/src/lib/components/toggle-status/toggle-status.component.ts
```diff
@@ -0,0 +1,18 @@
...
+import { TranslateModule } from '@ngx-translate/core';
+import { ButtonComponent } from '../button';
+
+@Component({
+  selector: 'asset-sg-toggle-status',
+  templateUrl: './toggle-status.component.html',
+  styleUrls: ['./toggle-status.component.scss'],
+  imports: [ButtonComponent, TranslateModule],
+})
+export class ToggleStatusComponent {
```
daniel-va, line 11: Here, too, you could employ [ControlValueAccessor](https://angular.dev/api/forms/ControlValueAccessor).

libs/client-shared/src/lib/components/toggle-status/toggle-status.component.html
```diff
@@ -0,0 +1,8 @@
+<div class="toggle">
```
daniel-va, line 1: This wrapper is not necessary, you can just style the `:host`.

libs/client-shared/src/lib/components/toggle-status/toggle-status.component.scss
```diff
@@ -0,0 +1,24 @@
...
+    transition: background-color 0.3s;
+  }
+
+  button:first-of-type {
+    margin-left: 6px;
+  }
+
+  button:last-of-type {
+    margin-right: 6px;
+  }
```
daniel-va, line 23: Isn't this just a roundabout way of giving `.toggle` a `padding-inline: 6px`?

daniel-va (changes requested):

libs/client-shared/src/lib/components/text-input/text-input.component.ts
```diff
@@ -1,22 +1,52 @@
...
+    },
+  ],
 })
-export class TextInputComponent {
+export class TextInputComponent implements ControlValueAccessor {
   @Input() public term = '';
   @Input() public icon = '';
   @Input() public placeholder = '';
-  @Output() termChanged = new EventEmitter<string>();
+  @Output() changed = new EventEmitter<string>();
```
daniel-va, line 25: The default HTML event for changes in input is called `change` - renaming this would align the component with that.

libs/client-shared/src/lib/components/text-input/text-input.component.ts
```diff
@@ -1,22 +1,52 @@
...
+    {
+      provide: NG_VALUE_ACCESSOR,
+      useExisting: TextInputComponent,
+      multi: true,
+    },
+  ],
 })
-export class TextInputComponent {
+export class TextInputComponent implements ControlValueAccessor {
   @Input() public term = '';
```
daniel-va, line 22: Very minor thing, but I would simply call this `value` - `term` is a remnant from back when this was a search input only, and does now seem slightly confusing.

libs/client-shared/src/lib/components/select/select.component.ts
```diff
@@ -0,0 +1,48 @@
...
+    ReactiveFormsModule,
+    SvgIconComponent,
+    TranslateModule,
+    FormsModule,
+    MatFormField,
+    SmartTranslatePipe,
+    MatHint,
+  ],
+})
+export class SelectComponent<T extends Translation> implements OnInit {
```
daniel-va: I like this being generic overall. The problem with the current implementation is that now the select only allows values that are themselves translatable. Ideally, I could use whatever I want for `T`, but then choose how that value is displayed.

Taking insipration from [`ng-select`](https://www.npmjs.com/package/@ng-select/ng-select) (which is a great select library IMO), you could add a `bindLabel: keyof T` input, which you then use retrieve the display value of each label via `value[bindLabel]`. This would allow you to drop the `Translation` constraint here, and even make it possible to use entire objects (i.e. `User`, `Workgroup`) as the select's values.

libs/client-shared/src/lib/components/filter-selector/filter-selector.component.ts
```diff
@@ -36,18 +38,27 @@ export class FilterSelectorComponent<T> implements OnInit {
   @Input() public initialValues: Array<T> = [];
   @Input() public shouldShowError = false;
   @Input() public errorMessage = '';
-  @Output() public filterChanged = new EventEmitter<Filter<T>[]>();
+  @Output() public filterAdded = new EventEmitter<Filter<T>[]>();
+  @Output() public filterRemoved = new EventEmitter<Filter<T>[]>();
```
daniel-va, line 29: I would recommend changing both of these outputs to emit single values (`EventEmitter<Filter<T>>`). This allows you to move the loop required to add/remove filters into this component, rather than having to repeat it at each use side.

libs/admin/src/lib/components/workgroups/workgroups.component.ts
```diff
@@ -16,18 +17,19 @@ type UsersPerRole = Array<[Role, number]>;
...
+export class WorkgroupsComponent extends AbstractAdminTableComponent<Workgroup> implements OnInit, OnDestroy {
   protected readonly COLUMNS = ['name', 'numberOfAssets', 'users', 'status', 'actions'];
   public workgroups: Array<Workgroup & { usersPerRole: UsersPerRole }> = [];
   public users: User[] = [];
 
   private readonly store = inject(Store<AppStateWithAdmin>);
-  public isActiveFilterValues = [
-    { displayValue: { key: 'admin.workgroupPage.isActive' }, value: true },
-    { displayValue: { key: 'admin.workgroupPage.isDisabled' }, value: false },
+  public isActiveFilterValues: Filter<Workgroup>[] = [
```
daniel-va, line 26: This name reads very weird to me, as my mind immediately goes `that's a boolean` due to the `is` prefix. I would rename this to something like `filtersForIsActive`.

libs/admin/src/lib/components/users/users.component.ts
```diff
@@ -18,21 +17,18 @@ import { AbstractAdminTableComponent } from '../abstract-admin-table/abstract-ad
...
 
   public workgroups = new Map<WorkgroupId, Workgroup>();
-  public workgroupFilterValues: Filter<number>[] = [];
-  public readonly langFilterValues: Filter<Lang>[] = [
-    { displayValue: 'DE', value: 'de' },
-    { displayValue: 'EN', value: 'en' },
-    { displayValue: 'FR', value: 'fr' },
-    { displayValue: 'IT', value: 'en' },
+  public workgroupFilterValues: Filter<User>[] = [];
+  public readonly langFilterValues: Filter<User>[] = [
```
daniel-va, line 25: Why don't you just drop the `Values` suffix for these variables? Does it have a special meaning for you?

libs/client-shared/src/lib/components/filter-selector/filter-selector.component.ts
```diff
@@ -19,46 +18,33 @@ export interface Filter<T> {
...
-    FormsModule,
-    MatFormField,
-    SmartTranslatePipe,
-    MatHint,
-  ],
+  imports: [ReactiveFormsModule, TranslateModule, FormsModule, SelectComponent],
 })
 export class FilterSelectorComponent<T> implements OnInit {
-  @Input() public values: Filter<T>[] = [];
+  @Input() public filterValues: Filter<T>[] = [];
```
daniel-va, line 24: I'd just call these `filters`.

libs/client-shared/src/lib/components/text-input/text-input.component.ts
TIL-EBP: When renaming to `change` i get a warning that i should not name events the same as standard DOM events

daniel-va: ah right, the Angular standards have something against that for some reason. Thinking about it now, it would make more sense anyways to rename the output to `termChange` (or whatever you end up calling your value input) so we can use the shorthand `[(term)]="myLocalValue"`.

libs/admin/src/lib/state/admin.effects.ts
```diff
@@ -99,6 +100,14 @@ export class AdminEffects {
     )
   );
 
+  public loadUser$ = createEffect(() =>
+    this.actions$.pipe(
+      ofType(actions.setUser, actions.setWorkgroup),
+      switchMap(() => this.authService.getUserProfile()),
```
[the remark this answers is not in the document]
TIL-EBP: We load it once when we start the app, on logout, and when we edit a user or workgroup. I think that's fine. We could restrict onyl oading the user when the current user or any of its associated workgroups has been updated, but tbh i couldnt be bothered

libs/client-shared/src/lib/components/filter-selector/filter-selector.component.html
```diff
@@ -1,15 +1,8 @@
...
-    @for (value of values; track value) {
-    <mat-option [value]="value">{{ value.displayValue | smartTranslate }}</mat-option>
-    }
-  </mat-select>
-  @if (shouldShowError) {
-  <mat-hint class="mat-mdc-form-field-error">{{ errorMessage }}</mat-hint>
-  }
-</mat-form-field>
+<asset-sg-select
+  [values]="values"
```
daniel-va, line 2: You are using the `displayValue` of the filters as key in the select, which seems weird to me. I would not expect a value meant for display be used as a unique key.

Why don't you just pass teh filters directly to the `asset-sg-select` and use `bindLabel` on `displayValue`?

libs/client-shared/src/lib/components/filter-selector/filter-selector.component.ts
```diff
@@ -17,37 +18,37 @@ export interface Filter<T> {
...
+      selectedValues = [selectedValues];
+    }
+    const added = this.filters.filter((value) => selectedValues.includes(value.displayValue));
+    const removed = this.filters.filter((value) => !selectedValues.includes(value.displayValue));
+    for (const filter of added) {
+      this.filterAdded.emit(filter);
+    }
+    for (const filter of removed) {
+      this.filterRemoved.emit(filter);
+    }
```
daniel-va, line 52: This is a bit inefficent - these are four loops, two of which contain the same nested loop.
This could be rewritten to something like this:

```ts
const newFilters = new Set(selectedValues);
for (const filter of this.filters) {
  const isNew = newFilters.delete(filter);
  if (!isNew) {
    this.filterRemoved.emit(filter);
  }
}
for (const newFilter of newFilters) {
    this.filterAdded.emit(filter);
}
```
\+ I don't think that you update `this.filters` anywhere?

libs/client-shared/src/lib/components/filter-selector/filter-selector.component.ts
```diff
@@ -17,37 +18,37 @@ export interface Filter<T> {
...
+  @Input() public initialValues: Array<Translation> = [];
+  @Output() public filterAdded = new EventEmitter<Filter<T>>();
+  @Output() public filterRemoved = new EventEmitter<Filter<T>>();
 
+  public values: Translation[] = [];
   public selectedFilters?: Filter<T> | Filter<T>[] = this.multiple ? [] : undefined;
 
   public ngOnInit(): void {
+    this.values = this.filters.map((filter) => filter.displayValue);
     if (this.initialValues.length > 0) {
```
daniel-va, line 36: This `if` is kind of unnecessary - it is simply a micro-optimization, no? You could just leave it out and nothing would change.

daniel-va (approved):

libs/client-shared/src/lib/components/select/select.component.ts
```diff
@@ -0,0 +1,79 @@
...
+  @Input() public errorMessage = '';
+  @Output() public selectionChanged = new EventEmitter<T[]>();
+
+  public selectedValues?: T | T[] = this.multiple ? [] : undefined;
+
+  private onChange: (value: T | T[]) => void = noop;
+  private onTouched: () => void = noop;
+
+  public ngOnInit(): void {
+    if (this.initialValues.length > 0) {
```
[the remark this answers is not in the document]
TIL-EBP: If there aren't any initial values set, we do not need to loop over the values themselves, which saves a loop. 
Won't make much of a difference but still prevents useless operations from happening

libs/client-shared/src/lib/components/filter-selector/filter-selector.component.ts
TIL-EBP: `this.filters` is the input I get, so it should not be updated. BUt you are right. I did not update `this.selectedFilters`

libs/client-shared/src/lib/components/filter-selector/filter-selector.component.ts
TIL-EBP: same here, the if saves us a loop over the filters

PR #429, opened 2025-02-24
Feature/assets 297 delta UI ux

### LINKED ISSUE #297: Delta UI/UX Anpassungen
**Beschreibung**

Aus einem Abgleich der Applikationen Assets und Boreholes wurden die folgenden Änderungen erarbeitet:

- Header
  - Kein Schatten unterhalb Header
  - Border-Bottom
  - Sprachen Dropdown rechts anschlagen
- Tabelle
  - Tabellen-Header verkleinern
  - Tabellen Container-Header vergrössern
  - Container-Header Titel und Tabelle aneinander ausrichten
  - “Tabelle verbergen” zu Button machen
- Grösse des Logos an Figma angleichen
- Karten Kontrollen an Boreholes angleichen, als Button und die Grösse entsprechend

TIL-EBP:
resolves #297

### DIFF
--- a/libs/client-shared/src/lib/components/button/button.component.scss
+++ b/libs/client-shared/src/lib/components/button/button.component.scss
@@ -114,6 +114,13 @@
   width: 28px;
 }
 
+:host([isMapControl]) {
+  padding: 0;
+  height: 36px;
+  width: 36px;
+  margin: 4px;
+}
+
 :host[asset-sg-icon-button] {
   @extend %reset;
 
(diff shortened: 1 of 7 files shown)


libs/client-shared/src/lib/components/button/button.component.scss
```diff
@@ -114,6 +114,13 @@
   width: 28px;
 }
 
+:host([isMapControl]) {
```
daniel-va, line 117: HTML attributes are normally written as `kebap-case` (i.e. `map-control`). I find it a bit weird to use an attribute here that is not defined as an input on the component itself.

Why is this style defined on `button.component` anyways? The style is only used for the map controls, as far as I can see.

Also, this most certainly should be a class, not an attribute.

daniel-va (approved):

daniel-va:
Were these already solved before this PR?

image

TIL-EBP:
Yes, all the header stuff had already been fixed

PR #430, opened 2025-02-24
replace break-word with overflow-wrap

### LINKED ISSUE #416: Bug: Redundanter Zeilenumbruch in der Mitte von Wörtern im Titel des Assets
Asset-Titel werden meistens an einer beliebigen Stelle in der Mitte von Wörtern abgebrochen für einen Zeilenumbruch:

Image

Dieses Verhalten scheint zurückzuführen zu sein auf die Lösung für https://github.com/swisstopo/swissgeol-assets-suite/issues/32 . Hier wurde eine CSS-Regel `word-break: break-all` hinzugefügt.

Ohne diese Regel wird der Titel wie gewünscht angezeigt:

Image

Für https://github.com/swisstopo/swissgeol-assets-suite/issues/32 wäre es wahrscheinlich eine bessere Lösung, die CSS-Regel `overflow-wrap: anywhere` zu benutzen. Gemäss https://developer.mozilla.org/de/docs/Web/CSS/overflow-wrap:

> Hinweis: Im Gegensatz zu [word-break](https://developer.mozilla.org/de/docs/Web/CSS/word-break) erzeugt overflow-wrap nur einen Umbruch, wenn ein ganzes Wort nicht ohne Überlauf in einer eigenen Zeile platziert werden kann.

Image

Image

TIL-EBP:
resolves #416

### DIFF
--- a/libs/asset-viewer/src/lib/components/asset-search-detail/asset-search-detail.component.html
+++ b/libs/asset-viewer/src/lib/components/asset-search-detail/asset-search-detail.component.html
@@ -17,7 +17,7 @@
     <div class="asset-detail">
       <ng-container *ngIf="assetDetail$ | push as assetDetail">
         <div class="flex flex-row justify-between mb-4">
-          <div class="text-dark-red break-word font-bold mt-2.5" [innerHTML]="assetDetail.titlePublic"></div>
+          <div class="text-dark-red font-bold mt-2.5" [innerHTML]="assetDetail.titlePublic"></div>
           <ng-container *canUpdate="AssetEditPolicy; with: assetDetail">
             <a asset-sg-icon-button [routerLink]="['/', currentLang$ | async, 'asset-admin', assetDetail.assetId]">
               <svg-icon key="edit" data-testid="edit-asset-button"></svg-icon>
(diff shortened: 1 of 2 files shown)


libs/asset-viewer/src/lib/components/asset-search-detail/asset-search-detail.component.html
```diff
@@ -17,7 +17,7 @@
     <div class="asset-detail">
       <ng-container *ngIf="assetDetail$ | push as assetDetail">
         <div class="flex flex-row justify-between mb-4">
-          <div class="text-dark-red break-word font-bold mt-2.5" [innerHTML]="assetDetail.titlePublic"></div>
+          <div class="text-dark-red font-bold mt-2.5" [innerHTML]="assetDetail.titlePublic"></div>
```
daniel-va, line 20: I know this is not part of this PR, but why is this an `innerHTML`? Can asset titles actually contain HTML?

daniel-va (approved):

libs/asset-viewer/src/lib/components/asset-search-detail/asset-search-detail.component.html
TIL-EBP: Good question. I haven't at all looked at the code except for the css, so I didn't even realize. 
I don't know where a user would specify a title with html, but I would'nt be surprised if there are some that have cursive text or something like that. I will try and create one and see what happens

daniel-va: Okay, thank you. If possible I'd like to have this removed, as it probably carries some security implications.

PR #434, opened 2025-03-05
Adjust readme with additional first setup information

Tugark:
Adds the missing bits for setting up the project as a new developer.


README.md
```diff
@@ -50,6 +50,14 @@ npm run prisma -- generate
   - Set `S3_ACCESS_KEY_ID` to your generated access key.
   - Set `S3_SECRET_ACCESS_KEY` to your generated access key's secret.
 
+#### 4. Define Admin User
+
+Per default, our user is not an admin and cannot access admin functionality. In order to so, manually set the user to admin in the database directly:
+```sql
+-- Note: this assumes that you
```
daniel-va, line 57: I guess this note is incomplete?

README.md
```diff
@@ -50,6 +50,14 @@ npm run prisma -- generate
   - Set `S3_ACCESS_KEY_ID` to your generated access key.
   - Set `S3_SECRET_ACCESS_KEY` to your generated access key's secret.
 
+#### 4. Define Admin User
+
+Per default, our user is not an admin and cannot access admin functionality. In order to so, manually set the user to admin in the database directly:
+```sql
+-- Note: this assumes that you 
+UPDATE asset_user SET is_admin = true WHERE email = 'admin@assets.swissgeol.ch';
```
daniel-va, line 58: I'd prefer to use a cli command here, so it - like all other commands in the README - can be run directly in the console.
```suggestion
docker compose exec db sh -c 'psql --dbname=postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@localhost:5432/${POSTGRES_DB} -c "UPDATE asset_user SET is_admin = true WHERE email = '\''admin@assets.swissgeol.ch'\''"'
```
Also, this command will only work if the user has signed in at least once, which we should probably mention here.

daniel-va (changes requested):

README.md
Tugark: Fine by me; also added the note about having to sign in once.

daniel-va (approved):

PR #449, opened 2025-03-13
Fix bug that displayed the wrong language for users

### LINKED ISSUE #447: Bug:  Sprache in User Detailansicht inkonsistent
**Beschreibung**
Unter der Benutzerverwaltung Detailansicht wird die Sprache nicht immer korrekt wiedergegeben.

Gilt für INT und DEV

1. Gehe zur Seite `/admin/users`
2. Klicke auf einen User mit z.B. Sprache `de`
3. Klicke zurück
4. Klicke auf einen anderen Benutzer mit anderer Sprache, z.B. `it`
5. Das Dropdown zeigt die erste Sprache, z.B. `deutsch` anstatt `italienisch`

**Beobachtetes Verhalten**
Das Sprachdropdown ist mit vorigen Werten befüllt.

**Erwartetes Verhalten**
Es werden die tatsächlichen Daten des aktuellen Benutzers gezeigt.

TIL-EBP:
resolves #447

### DIFF
--- a/libs/admin/src/lib/components/admin-page/admin-page.component.ts
+++ b/libs/admin/src/lib/components/admin-page/admin-page.component.ts
@@ -18,8 +18,8 @@ import { selectIsLoading, selectSelectedUser, selectSelectedWorkgroup } from '..
 export class AdminPageComponent implements OnInit, OnDestroy {
   @ViewChild('templateDrawerPortalContent') templateDrawerPortalContent!: TemplateRef<unknown>;
 
-  public workgroup?: Workgroup = undefined;
-  public user?: User = undefined;
+  public workgroup: Workgroup | null = null;
+  public user: User | null = null;
   private readonly store = inject(Store<AppState>);
   public readonly isLoading$ = this.store.select(selectIsLoading);
   public readonly selectedUser$ = this.store.select(selectSelectedUser);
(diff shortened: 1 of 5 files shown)


libs/admin/src/lib/components/admin-page/admin-page.component.ts
```diff
@@ -19,6 +19,7 @@ export class AdminPageComponent implements OnInit, OnDestroy {
   @ViewChild('templateDrawerPortalContent') templateDrawerPortalContent!: TemplateRef<unknown>;
 
   public workgroup?: Workgroup = undefined;
+  public user?: User = undefined;
```
Tugark, line 22: Not sure about coding conventions - but I'd remove the explicit `undefined` definition since it is implied by using `user?`.

libs/admin/src/lib/components/admin-page/admin-page.component.ts
```diff
@@ -91,5 +92,11 @@ export class AdminPageComponent implements OnInit, OnDestroy {
         this.workgroup = workgroup ?? undefined;
       })
     );
+
+    this.subscriptions.add(
+      this.selectedUser$.subscribe((user) => {
+        this.user = user ?? undefined;
```
Tugark, line 98: Is there a reason that `selectSelectedUser` returns `null`?

I think you could simplify it if that one returned `undefined` instead of `null`; or then I'd question why `this.user` is `undefined` and needs this explicit mapping from `null` to `unfedined`

TIL-EBP: Fair point. I honestly do not know why we defined it that way, but in the state the user is defined as `User | null` (same for workgroups). I guess this could be changed to `User | undefined`, but I don't really feel like touching the state now for this bug. 
So i just changed the `this.user` to be `User | null` instead

Tugark (approved):

PR #457, opened 2025-03-18
Update dependencies

daniel-va:


libs/asset-editor/src/lib/components/asset-editor-id-form/asset-editor-id-form.component.ts
```diff
@@ -27,7 +27,7 @@ export class AssetEditorIdFormComponent implements OnInit, OnChanges, OnDestroy
   public delete = new EventEmitter<IdVM>();
 
   @Output()
-  public finish = new EventEmitter<void>();
+  public closeForm = new EventEmitter<void>();
```
TIL-EBP, line 30: Why add this change here? 
The build fails now because we still reference `this.finish` in the component (line 94)

TIL-EBP (changes requested):
Build fails due to property that no longer exists

libs/asset-editor/src/lib/components/asset-editor-id-form/asset-editor-id-form.component.ts
daniel-va: The upgrade updated the eslint rule that checks that outputs can't be named the same as HTML events. Apparently, there is now an official `finish` event.

TIL-EBP (approved):

PR #472, opened 2025-03-26
Feature/assets 300 legend switch implementation

### LINKED ISSUE #300: Freigabe als alternative Legende
**Beschreibung**

Es soll eine weitere Palette geben, in der je nach Status der Freigabe eingefärbt wird. 

Image

**Acceptance Criteria**

- [ ] Die Legende ist funktional und die Features auf der Karte entsprechen den Farben der Legende.

Tugark:
I had to refactor quite some parts of how we deal with stylings. The main change is that I switched from feature-based styling to layer based styling using stylefunctions. Adding a new layer always requires defining a new style function; and this function is applied to the asset/study layers and changed when toggling the style.

As for the highlight - this required some more changes; essentially, we also apply the active style function to the layer where we add the chosen asset _and_ its buffered highlight geometry. Now, the fun thing that happens is because we add an explicit style to the buffer geometry (using `feature.setStyle()`) but _not_ to the copied geometry, the latter one is still styled like the normal layer, thus faking an override for the increased opacity on the main layer.

Note that it is _still_ not optimal and some parts are rather hacky; but in order to get it done, I tried to use most of the logic that is available. There are also some race conditions, which requires us to work around (e.g. creating a "new" study object if the assets are not yet loaded).

Also includes some minor refactorings, such as using an `Enum` for defining the property keys on the untyped `ol/Feature` properties.

**Note:** I already merged the changes of daniel into the `feature/legend-style-switch` branch, and merged them into here so we don't have any conflicts. However, I did _not_ add the style to the URL as well; I think this can be done in a later ticket.

Resolves #300


libs/asset-viewer/src/lib/shared/map-configuration/styles/access-based.map-layer-style.ts
```diff
@@ -0,0 +1,177 @@
...
+  if (!geometry) {
+    return new Style();
+  }
+
+  const accessType = feature.get(CustomFeatureProperties.AccessType) as StudyAccessType;
+  const accessStyles: keyof AccessTypeKey = accessTypeMapping[accessType];
+  switch (geometry.getType()) {
+    case 'Point': {
+      const geomType = feature.get(CustomFeatureProperties.GeometryType) as StudyGeometryType;
+      const styleKey = (
```
TIL-EBP, line 161: It took me a second to understand this. It might be easier to read if we have a mapping function from geomType to instanceType? Especially if new legends will be added in the future

TIL-EBP: Or maybe define the obbject below as an enum as well?

libs/asset-viewer/src/lib/components/map/map-controller.ts
```diff
@@ -611,30 +594,23 @@ const makeSources = (layers: MapLayers): MapLayerSources => ({
...
       }
     }
   })();
 
   const feature = new Feature({ geometry });
 
   feature.setId(study.studyId);
-  feature.setStyle(style);
-  feature.setProperties({ 'swisstopo.type': 'StudyGeometry' });
+  feature.setProperties({ [CustomFeatureProperties.SwisstopoType]: 'StudyGeometry' });
```
TIL-EBP, line 613: what is the swisstopType property used for?

TIL-EBP (approved):

libs/asset-viewer/src/lib/components/map/map-controller.ts
Tugark: I actually don't know :D But since it's been in the code before, I left it. But yeah, I'll remove it and see wheter it burns :D

libs/asset-viewer/src/lib/shared/map-configuration/styles/access-based.map-layer-style.ts
Tugark: Not sure what you mean by "define as an enum", but yeah, I agree, this was hard to read. I extracted the lookup table (similar to the `accessTypeMapping` constant, and created a function that performs the lookup. 

It's still _somewhat_ hard to read, but it is confined to this style. By the way - adding other legends does not have an influence, since this style here is isolated for this mapstyle only.

TIL-EBP: What I meant by "new legends will be added" is that for every legend style that styles the faetures by property, we will have to map the geomType to the geomInstance, so this snipped will be reused in new legends as well

Tugark: @TIL-EBP Fair point. But given that this should be as typesafe as possible, it's not a simple enum. Instead, I added a mapper function, does this make sense to you?

TIL-EBP: Yeah I realized that an enum would not work in this case, so the mapper function is perfect.

PR #488, opened 2025-04-01
Merge hotfixes into `develop`

daniel-va:


.github/workflows/publish-edge.yml
```diff
@@ -5,6 +5,9 @@ on:
     branches:
       - "develop"
 
+      # TODO remove before merge
+      - "feature/assets-437-ermoegliche-permalinks"
```
Tugark, line 9: Do we remove this now, or when do you remove it?

.github/workflows/publish-edge.yml
```diff
@@ -5,6 +5,9 @@ on:
     branches:
       - "develop"
 
+      # TODO remove before merge
```
TIL-EBP, line 8: commenting for visiblity :D

TIL-EBP (approved):

PR #508, opened 2025-04-03
Assets-492: Database Schema and CRUD for workflows

### LINKED ISSUE #492: Workflows Database & API CRUD
**Beschreibung**

Implementiere die Anpassungen an Datenbank und API, damit Workflows möglich sind entsprechend technischer Planung.

Image

Spezielle Punkte:
* Asset creation -> muss Workflow objekt anlegen (1:1)
* Data Migration -> bestehende Assets müssen einen (default) erhalten
* Evtl Business Rules (z.B. skippen von Status, etc.)
* 

**Abgrenzung**

Die neuen Workgroup-Rollen sowie Authorisierung sind nicht teil dieser Aufgabe.

**Acceptance Criteria**

- [ ] Per API können Workflows benutzt werden.

Tugark:
Partially resolves #492 

Unfortunately, I wasn't able to implement all the endpoints, so some things to note:

* I added `@Modules` to better structure the related concepts, and to make importing across modules easier. Also, once we get rid off `asset-edit`, we can add global route prefixes in our `RouterModule` so we do not have to repeat the paths in the route decorators
* Contrary to our specification, I removed the `isPublished` flag and added `published` as an object state - this is to streamline what we do in the frontend, since the publication is used similar to other states (i.e. it has an entry, it is set on the workflow, etc.)
* There is quite some logic going on (of which not all are implemented, see below); hence I added a servicelayer instead of doing it in the repo or controllers. Given that we might have even more checks and validations, I thought that this would be the best idea without blowing up our controllers. Also note that we need to discuss how we handle status resets, e.g. when someone starts editing an asset - this does not explicitly call our workflow endpoints.
* Tests fail, but I couldn't reproduce it locally because my testsetup does not work.
* As per the design - it is not "restful" per se, but I think this is fine because we always only want to work on subsets of the workflow (e.g. adding a change does not affect any other params; publishing a workflow does not affect other params), so I added "non RESTful routes". 
* The PATCH route for Review and Published tabs is not yet implemented, sorry. However, that one should be straightforward PATCH and just affect the fields that change.

Note that there is a lot of edge cases that we will need to handle. I think they can be handled best once we start developing the frontend, since it's quite hard to think of all of them right now. This PR should set, however, an appropriate base for us to work on.


libs/persistence/prisma/schema.prisma
```diff
@@ -682,3 +688,70 @@ enum StudyGeometryType {
...
+  InReview
+  Reviewed
+
+  @@map("workflow_status")
+}
+
+model TabStatus {
+  tabStatusId Int @id @default(autoincrement()) @map("tab_status_id")
+
+  general    Boolean @default(false)
```
TIL-EBP, line 746: Are these the definitive tabs now? or will they add "subcategories"?

TIL-EBP (approved):

libs/persistence/prisma/schema.prisma
Tugark: We'll have to wait and see. I'll change the migration once we know the definitive ones.

apps/server-asset-sg/src/features/assets/workflow/workflow.http
```diff
@@ -0,0 +1,19 @@
...
+Authorization: Impersonate {{user}}
+Content-Type: application/json
+
+{
+  "assignee": "admin@assets.swissgeol.ch",
+  "comment": "Test comment",
+  "toStatus": "reviewed"
+}
+
+### Add change to a workflow
```
TIL-EBP, line 16:
```suggestion
### Publish a workflow
```

TIL-EBP (approved):

PR #512, opened 2025-04-10
Feature/assets 500 edit page

### LINKED ISSUE #500: Asset-Edit Page, Form, Tabs erstellen
**Beschreibung**

Für die neue Implementierung der Detailansicht Assets, erstelle eine neue Seite inklusive Form und der Tabstruktur entsprechend Figma.
Es soll weiterhin ein Modal beim Verlassen mit ungespeicherten Änderungen erscheinen.
Die neue Struktur, mit dem `Abbrechen` und `Speichern` Button als sticky footer soll auch implementiert sein.

Allenfalls ein einzelnes Input wie z.B. Title hinzufügen, damit man das Speicher-Verhalten testen kann.

**Abgrenzung**

Der Inhalt der einzelnen Tabs ist nicht Teil dieses Issues. 

**Acceptance Criteria**

- [ ] Die neue Asset-Edit page existiert

TIL-EBP:
resolves #500 
Sets up the form and ui for edit copmonent.
I also moeved some things around, most notably the `currentAsset` is no in the shared State. I am still unsure if this was the correct desicion. If you don't like this change, feel free to tell me and i will revert it. 
I also removed some of the fp-ts used in the effects and service for the edit, but some is still there as i did not touch all methods


libs/asset-editor/src/lib/components/asset-editor-page/asset-editor-page.component.ts
```diff
@@ -1,50 +1,143 @@
...
+  private readonly subscriptions: Subscription = new Subscription();
+
+  readonly activeItem$ = this.router.events.pipe(
+    filter((event) => event instanceof NavigationEnd),
+    startWith(() => undefined),
+    tap(() => {
+      const segments = (this.router.getCurrentNavigation() ?? this.router.lastSuccessfulNavigation)?.finalUrl?.root
+        .children?.['primary']?.segments;
+      if (segments !== undefined && segments.length > 0) {
+        this.activeTab = segments.pop()?.path as Tab;
```
daniel-va, line 42: I don't see the need of using `.pop()` over `[segments.length - 1]` here - it is it even safe to modify the segments array?

libs/asset-editor/src/lib/components/asset-editor-page/asset-editor-page.component.ts
```diff
@@ -1,50 +1,143 @@
...
+      }
+    })
+  );
+
+  public ngOnInit() {
+    this.subscriptions.add(this.activeItem$.subscribe());
```
daniel-va, line 48: okay, this just looks weird - is `activeItem$` not actually used anywhere, and is instead just used as an effect? If so, I would strongly prefer fully moving it into a method insead of keeping it around as an attribute.

libs/asset-editor/src/lib/components/asset-editor-page/asset-editor-page.component.ts
```diff
@@ -1,50 +1,143 @@
...
+    this.subscriptions.unsubscribe();
+  }
+
+  public navigateToStart() {
+    this.routingService.navigateToRoot().then();
+  }
+
+  public initializeForm() {
+    this.form.reset();
+    this.form.controls.general.controls.titlePublic.setValue(this.asset?.titlePublic ?? null);
```
daniel-va, line 71: Is the end goal here to set every one of the asset's attributes to one of the form's fields?

libs/asset-editor/src/lib/components/asset-editor-page/asset-editor-page.component.ts
```diff
@@ -1,50 +1,143 @@
...
+  public asset: AssetEditDetail | null = null;
+  public form!: AssetForm;
+  public activeTab: Tab = Tab.General;
+  private readonly store = inject(Store<AppSharedState>);
+  private readonly route = inject(ActivatedRoute);
+  private readonly router = inject(Router);
+  private readonly routingService = inject(RoutingService);
+  private readonly dialogService = inject(MatDialog);
+
+  public assetEditDetail$ = this.store.select(fromAppShared.selectCurrentAsset);
```
daniel-va, line 31:
```suggestion
  private readonly assetEditDetail$ = this.store.select(fromAppShared.selectCurrentAsset);
```

libs/asset-editor/src/lib/components/asset-editor-page/asset-editor-page.component.ts
```diff
@@ -1,50 +1,143 @@
...
+  private loadAssetFromRouteParams() {
+    this.subscriptions.add(
+      this.route.paramMap
+        .pipe(withLatestFrom(this.store.select(fromAppShared.selectCurrentAsset)), take(1))
+        .subscribe(([params, currentAsset]) => {
+          const assetIdFromParam = params.get('assetId');
+          if (!assetIdFromParam || !parseInt(assetIdFromParam)) {
+            return;
+          }
+          const assetId = parseInt(assetIdFromParam);
```
daniel-va, line 111: You're using `parseInt` twice - I'd prefer to just drop the preceding parse step and instead do an `isNaN(assetId)` check afterward.

libs/asset-editor/src/lib/components/asset-editor-page/asset-editor-page.component.ts
```diff
@@ -1,50 +1,143 @@
...
-  constructor() {
-    this.lc.afterViewInit$.pipe(observeOn(asyncScheduler)).subscribe(() => {
-      this.appPortalService.setAppBarPortalContent(null);
-      this.appPortalService.setDrawerPortalContent(
-        new TemplatePortal(this._templateDrawerPortalContent, this.viewContainerRef)
-      );
-      this.cd.detectChanges();
+export class AssetEditorPageComponent implements OnInit, OnDestroy {
+  public assetId: number | string = 'new';
+  public asset: AssetEditDetail | null = null;
```
daniel-va, line 22: Is this `null` when this form is used to create a new asset? If so, I'd like to see a comment explaining this. I'm also currently a bit confused about the relationshiop of `asset` and `assetId`.

libs/asset-editor/src/lib/components/asset-editor-page/asset-editor-page.component.ts
```diff
@@ -1,50 +1,143 @@
...
+            this.store.dispatch(actions.loadAsset({ assetId }));
+          }
+        })
+    );
   }
+
+  protected readonly Tab = Tab;
 }
+
+function buildForm() {
```
daniel-va, line 127: Arrow functions > regular functions 😊 
```suggestion
const buildForm = () => {
```

libs/asset-editor/src/lib/components/asset-editor-save/asset-editor-save.component.html
```diff
@@ -0,0 +1,15 @@
+@if (hasUnsavedChanges) {
+<div class="warning">
+  <svg-icon key="cancel"></svg-icon>
+  <span translate>edit.unsavedChanges</span>
+</div>
+}
+<div style="flex: 1"></div>
```
daniel-va, line 7: Make a class.

libs/asset-editor/src/lib/state/asset-editor.effects.ts
```diff
@@ -86,11 +36,11 @@ export class AssetEditorEffects {
...
       switchMap(({ assetId, patchAsset, newFiles, filesToDelete }) =>
         this._assetEditorService.deleteFiles(assetId, filesToDelete).pipe(
-          ORD.chainSwitchMapW(() => this._assetEditorService.uploadFiles(assetId, newFiles)),
-          ORD.chainSwitchMapW(() => this._assetEditorService.updateAssetDetail(assetId, patchAsset))
+          switchMap(() => this._assetEditorService.uploadFiles(assetId, newFiles)),
+          switchMap(() => this._assetEditorService.updateAssetDetail(assetId, patchAsset))
         )
       ),
-      map((data) => actions.updateAssetEditDetailResult({ data }))
+      map((data) => actions.updateAssetEditDetailResult({ asset: data }))
```
daniel-va, line 43: Could just do this:
```suggestion
      map((asset) => actions.updateAssetEditDetailResult({ asset }))
```

libs/asset-viewer/src/lib/services/viewer-params.service.ts
```diff
@@ -20,13 +21,16 @@ export class ViewerParamsService {
   private readonly store = inject(Store<AppStateWithAssetSearch>);
 
   async readParamsFromStore(): Promise<ViewerParams> {
-    const state = await firstValueFrom(
+    const searchState = await firstValueFrom(
       this.store.pipe(map((store) => store.assetSearch)) as Observable<AssetSearchState>
     );
+    const sharedState = await firstValueFrom(
+      this.store.pipe(map((store) => store.shared)) as Observable<AppSharedState>
+    );
```
daniel-va, line 29: Probably a good idea to unify these.
```suggestion
    const [searchState, sharedState] = await firstValueFrom(
      this.store.pipe(map((store) => [store.assetSearch, store.shared] as [AssetSearchState, AppSharedState]))
    );
```

daniel-va (changes requested):
Looks like we're going in a good direction - there are still some parts that are unclear/confusing to me, which is something i'd like to have ironed out before we're moving on to the next step of the implementation.

Overall, I really like moving away from the old form's structure, and very happy to get rid of the old stuff, like the RemoteData.

libs/asset-editor/src/lib/components/asset-editor-page/asset-editor-page.component.html
```diff
@@ -1,7 +1,37 @@
-<ng-template #templateDrawerPortalContent>
-  <asset-sg-drawer #tabPageComponent>
-    <asset-sg-drawer-panel widthPercentageOfViewPanel="1" [display]="true">
-      <asset-sg-editor-tab-page></asset-sg-editor-tab-page>
-    </asset-sg-drawer-panel>
-  </asset-sg-drawer>
-</ng-template>
+<asset-sg-page-header (backNavigation)="navigateToStart()">
+  <h2>{{ asset?.titlePublic }}</h2>
```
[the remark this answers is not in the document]
TIL-EBP: Currently, yes. It is not defiend what we are placing there when creating a new asset as of now, so I didn't bother coming up with something

libs/asset-editor/src/lib/components/asset-editor-page/asset-editor-page.component.ts
TIL-EBP: Basically, yes. We need some place to fill the form. If there is another way that is easier, then thats great

libs/asset-editor/src/lib/components/asset-editor-page/asset-editor-page.component.ts
TIL-EBP: why? is this only for legibiility in this case? or are there other reasons?

libs/asset-editor/src/lib/components/asset-editor-navigation/index.ts
```diff
@@ -0,0 +1 @@
+export * from './asset-editor-navigation.component';
```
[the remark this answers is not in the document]
TIL-EBP: True. Hoestly I usually just copied the content of one component folder to create a new one (becuase `ng g c new-component` does not seem to work inside the libs. BUt i am happy the leave them out/ reomve them

libs/asset-editor/src/lib/components/asset-editor-page/asset-editor-page.component.ts
daniel-va: I'm not against doing this here, it was more for clarification so I understand what's going on 😄.

Another way I tend to do this is by giving the asset to `buildForm` and then just using an entire new form when resetting, e.g.
```ts
const buildForm = (asset: Asset | null) => new FormGroup({
  general: new FormGroup({
    titlePublic: new FormControl(asset?.titlePublic ?? ''),
    ...
  }),
  ...
});
```
This requires a bit less code and reduces the risk of forgetting to set a control. However, it brings its own set of problems, such as bricking form subscriptions.

libs/asset-editor/src/lib/components/asset-editor-page/asset-editor-page.component.ts
daniel-va: My main two reasons are for standardization and because I find them easier to reason about (among other things, regular functions are subject to [hoisting](https://developer.mozilla.org/en-US/docs/Glossary/Hoisting), which is also the reason you should use `let` over `var`).

Another thing is that `function`s have their own [`this`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this) binding, while arrow functions don't. This, to me, signals that whenever I see a `function`, that its `this` has a special meaning.

libs/asset-editor/src/lib/components/asset-editor-navigation/index.ts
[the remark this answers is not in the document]
daniel-va: I'd be glad if you could. Maybe we could also add a file to the repo, something like `CODE_STYLE`, where we keep track of decisions such as this, especially now that we are not just two people working on this project anymore.

libs/asset-editor/src/lib/components/asset-editor-page/asset-editor-page.component.ts
TIL-EBP: I removed `assetId` and replaed it with a `mode`-property to distinguish between creating and editing

TIL-EBP: The `currentAsset` in the state is also `null` if it is not set, so this is for consistency. It is possible that the asset is not loaded when the edit page is opened (on reload)

libs/asset-editor/src/lib/components/asset-editor-page/asset-editor-page.component.ts
```diff
@@ -1,50 +1,143 @@
...
+
+  public assetEditDetail$ = this.store.select(fromAppShared.selectCurrentAsset);
+
+  private readonly subscriptions: Subscription = new Subscription();
+
+  readonly activeItem$ = this.router.events.pipe(
+    filter((event) => event instanceof NavigationEnd),
+    startWith(() => undefined),
+    tap(() => {
+      const segments = (this.router.getCurrentNavigation() ?? this.router.lastSuccessfulNavigation)?.finalUrl?.root
```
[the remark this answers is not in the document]
TIL-EBP: I created a new injectionToken called `ROUTER_SEGMENTS`

CODE_STYLE.md
```diff
@@ -0,0 +1,7 @@
+# Code Style Guide
+
+This file collects the code style guidelines for the project. It is a work in progress and will be updated as needed.
+
+# Imports
```
daniel-va, line 5:
```suggestion
## Imports
```

daniel-va (approved):

PR #528, opened 2025-04-23
Feature/assets 506 admin tab

### LINKED ISSUE #506: Asset-Edit Tab Altdaten
**Beschreibung**

Implementiere den Altdaten Tab ensprechend Figma

Tugark:
Closes #506 

* Since the data is _never_ editable, I removed the form component for this and just use the hardcoded model
* I did some refactoring where the tabs are generated; mostly because before, that parts was fired twice (once on init, once when the asset was loaded) which seemed off.
* The logic for deciding whether we show the "Altdaten" tab is hardcoded and not generic; we can do it in a generic way should that requirement arise.
* I removeed `@UntilDestroyed` from your components @TIL-EBP as discussed
* I _think_ we could refactor some SCSS, notably the wraper around the page; because this is now duplicated across all pages. However, we could maybe do this at a later stage.
* Note: I asked Simon regarding the SGS-ID, because as I understood it, each asset will ALWAYS have an SGS-ID, so the tab will basically be visible all the time. Should this change, we can remove it.

### DIFF
--- a/libs/asset-editor/src/lib/components/asset-editor-page/asset-editor-page.component.ts
+++ b/libs/asset-editor/src/lib/components/asset-editor-page/asset-editor-page.component.ts
@@ -10,11 +10,11 @@ import {
   ROUTER_SEGMENTS,
   RoutingService,
 } from '@asset-sg/client-shared';
-import { AssetEditDetail, dateFromDateId, Lang } from '@asset-sg/shared';
+import { AssetEditDetail, dateFromDateId, hasHistoricalData, Lang } from '@asset-sg/shared';
 import { Workflow } from '@asset-sg/shared/v2';
 import { UntilDestroy, untilDestroyed } from '@ngneat/until-destroy';
 import { Store } from '@ngrx/store';
-import { Observable, Subscription, take, tap } from 'rxjs';
+import { EMPTY, filter, Observable, of, Subscription, switchMap, take, tap } from 'rxjs';
 import { EditorMode } from '../../models';
 import * as actions from '../../state/asset-editor.actions';
 import { selectWorkflow } from '../../state/asset-editor.selector';
@@ -39,9 +39,11 @@ export class AssetEditorPageComponent implements OnInit, OnDestroy {
    * The asset's workflow. This is `null` while the workflow is being loaded, or when a new asset is being created.
    */
   public workflow: Workflow | null = null;
-
   public form!: AssetForm;
   public activeTab: Tab = Tab.General;
+  protected availableTabs: Tab[] = [];
+  protected readonly Tab = Tab;
+  protected readonly EditorMode = EditorMode;
   private currentLang: Lang = 'de';
   private readonly store = inject(Store<AppSharedState>);
   private readonly route = inject(ActivatedRoute);
@@ -49,7 +51,6 @@ export class AssetEditorPageComponent implements OnInit, OnDestroy {
   private readonly dialogService = inject(MatDialog);
   private readonly currentLang$ = inject(CURRENT_LANG);
   private readonly routerSegments$ = inject(ROUTER_SEGMENTS);
-
   private readonly subscriptions: Subscription = new Subscription();
 
   public ngOnInit() {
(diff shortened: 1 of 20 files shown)


TIL-EBP (approved):

libs/asset-editor/src/lib/components/asset-editor-page/asset-editor-page.component.ts
```diff
@@ -69,12 +70,27 @@ export class AssetEditorPageComponent implements OnInit, OnDestroy {
...
+              );
+            } else {
+              this.initializeTabs();
+              this.initializeForm();
+              return EMPTY;
+            }
+          }),
+        )
+        .subscribe(),
     );
```
daniel-va, line 94: Any reason that this needs to be a subscription? As far as I can see you're now essentially running this sequentially, once, as the subscription uses the single-use `of(this.mode)` as source.

libs/asset-editor/src/lib/components/asset-editor-page/asset-editor-page.component.ts
```diff
@@ -141,6 +157,15 @@ export class AssetEditorPageComponent implements OnInit, OnDestroy {
     return dialogRef.afterClosed();
   }
 
+  private initializeTabs() {
+    this.availableTabs = Object.values(Tab).filter((tab) => {
+      if (tab === Tab.LegacyData) {
+        return !!(this.asset && hasHistoricalData(this.asset));
+      }
+      return true;
```
daniel-va, line 165:
```suggestion
      return tab !== Tab.LegacyData || !!(this.asset && hasHistoricalData(this.asset))
```

libs/shared/src/lib/models/asset-edit.ts
```diff
@@ -110,3 +110,8 @@ export const AssetEditDetail = C.struct({
   studies: C.array(C.struct({ assetId: C.number, studyId: C.string, geomText: C.string })),
 });
 export type AssetEditDetail = C.TypeOf<typeof AssetEditDetail>;
+
+// todo: once AssetEditDetail is removed, a similar property could exist on the new model to avoid this helper method
+export const hasHistoricalData = (asset: AssetEditDetail) => {
+  return asset.geolAuxDataInfo || asset.geolContactDataInfo || asset.geolDataInfo || asset.municipality || asset.sgsId;
+};
```
daniel-va, line 117: May be nice to make this type safe.
```suggestion
export const hasHistoricalData = (asset: AssetEditDetail): boolean => {
  return !!(asset.geolAuxDataInfo || asset.geolContactDataInfo || asset.geolDataInfo || asset.municipality || asset.sgsId);
};
```

libs/asset-editor/src/lib/components/asset-editor-page/asset-editor-page.component.ts
Tugark: It runs sequentially in that it returns a different observable depending on the value of `this.mode`. However, the subscription is required, because we need the returned observable:
* If we're not in edit mode, we can directly initialize everything and return an `EMPTY` observable - to which we subscribe, which is the immediately completed
* If we're in edit mode, we return an observable using `this.store.select`. Because we need to wait for it to be populated from the store, we filter it, and then use `take(1)` to have the observable complete automatically

I get that it's not that nice and rewriting it using signals would be simpler, but compared to before, we run the actual code only once in edit mode, where as it ran twice before (once for the initial value of `this.asset` and once for the loaded asset. Leaving it like this would have a flickering render in the navigation, because the check for the legacy data tab would only be inserted on the second run.

daniel-va: Ah right, I didn't catch that the `switchMap` actually returns an observable. Still wonder about the general structure of this statement - wouldn't this be much simpler when writing it like this:

```ts
if (mode === EditorMode.Edit) {
  this.subscription.add(this.store.select(fromAppShared.selectCurrentAsset).pipe(
    filter((asset) => !!asset),
    take(1),
    tap((asset) => {
      this.asset = asset;
      this.initializeTabs();
      this.initializeForm();
    }),
  ).subscribe());
} else {
  this.initializeTabs();
  this.initializeForm();
}
```

Tugark: I do like the RXJS approach better since it uses the same logic for both (and also implicitly handles the subscription logic) and it ties in neatly with how the other subscriptions are handled.

But our discussion shows that it might be in fact not _that_ readable, so I'll change it to your solution 👍

Tugark: Fixed.

PR #529, opened 2025-04-23
Add lint-staged

Tugark:


.husky/pre-commit
```diff
@@ -1 +1,3 @@
+echo "🚀 Running linter!"
 npx lint-staged
+echo ""
```
daniel-va, line 3: Is this newline intended?

daniel-va (approved):
Quick heads up, there is stil a `lint-staged` config in the `package.json`, which I don't think is needed (maybe it doesn't event do anything).

Otherwise nice & GTG.

.husky/pre-commit
Tugark: We had to add this in GB3, but I just checked and it seems to work without here; removed.

PR #536, opened 2025-04-30
Feature/assets 503 contacts editor

### LINKED ISSUE #503: Asset-Edit Tab Contacts
**Beschreibung**

Implementiere den Contacts Tab ensprechend Figma

Tugark:
Closes #503.

Adds contacts as per Figma's extended design (by reducing multiple roles into one, while still maintaining them as unique roles in the back).

I'm not _particularly_ happy with how the modals are handled, but it was the most straight forward solution and works quite nicely. I also tried to get rid of any io-ts stuff; however, I did not feel like rewriting the whole logic in the reducer, so I just merged my non-RD objects with the existing RD objects.

Also, as discussed, I no longer rely on effects.


libs/asset-editor/src/lib/components/asset-editor-tabs/asset-editor-contacts/asset-editor-contacts.component.ts
```diff
@@ -0,0 +1,152 @@
...
+  }
+
+  protected removeContact(element: ContactItem) {
+    this.form.controls.assetContacts.controls
+      .filter((control) => control.value.id === element.id)
+      .forEach((contact) => {
+        const idx = this.form.controls.assetContacts.controls.findIndex(
+          (control) => control.value.id === contact.value.id,
+        );
+        this.form.controls.assetContacts.removeAt(idx);
```
TIL-EBP, line 136: you are removing things from the array you are iterating over, which could potentially skip elements. 

Couldn't this whole methode be simplified by finding the index of the contact to delete and then removing it after? Why the filter?
can the same contact be present multiple times? I tried this and it just adds another role tag to the exisiting contact (even if it is the same)

TIL-EBP: Okay i understood now what is happening  with the filter, the same contact with multiple roles is stored as unique entires in the array. but still, modifying the array while looping over it seems dangerous

libs/client-shared/src/lib/components/select/select.component.html
```diff
@@ -13,7 +13,7 @@
       }
     </mat-select>
     @if (errorMessage) {
-      <mat-hint class="mat-mdc-form-field-error">{{ errorMessage }}</mat-hint>
+      <mat-hint class="mat-mdc-form-field-error">{{ errorMessage }}a</mat-hint>
```
TIL-EBP, line 16:
```suggestion
      <mat-hint class="mat-mdc-form-field-error">{{ errorMessage }}</mat-hint>
```

libs/asset-editor/src/lib/components/asset-editor-tabs/asset-editor-general/asset-editor-general.component.ts
```diff
@@ -98,22 +99,3 @@ export class AssetEditorGeneralComponent implements OnInit, OnDestroy {
     this.form.markAsDirty();
   }
 }
-
-export const mapValueItemsToTranslatedItem = (item: Record<string, ValueItem> | null): TranslatedValueItem[] => {
```
TIL-EBP, line 102: in on of my recent prs i also moved this somewehre else

libs/asset-editor/src/lib/components/asset-editor-tabs/asset-editor-contacts/link-contact-dialog/link-contact-dialog.component.ts
```diff
@@ -0,0 +1,112 @@
...
+          enterAnimationDuration: '0ms',
+          exitAnimationDuration: '0ms',
+          data: {
+            name: this.searchTerm ?? '',
+            roles: this.linkContactForm.value.roles,
+          },
+        })
+        .afterClosed()
+        .pipe(
+          tap((newContact) => {
```
TIL-EBP, line 101: would it be possible to immediately close this dialog and pass the reference of the newly opened dialog in the `close()` method, then then, in the `asset-editor-contacts-ts` component, within the subscription of the `afterClosed`, add another subscription tothe `afterClosed `of the passed reference? this way, there would be no flicker when closing the second dialog of the first dialog and you would need to pass the data through there anymore. Not sure if this even would work though (or if what i wrote makes any sense)

libs/asset-editor/src/lib/components/asset-editor-tabs/asset-editor-contacts/link-contact-dialog/link-contact-dialog.component.ts
```diff
@@ -0,0 +1,112 @@
...
+type SelectableContact = Pick<Contact, 'name' | 'id'>;
+
+@Component({
+  selector: 'asset-sg-link-contact-dialog',
+  templateUrl: './link-contact-dialog.component.html',
+  styleUrls: ['./link-contact-dialog.component.scss'],
+  standalone: false,
+})
+export class LinkContactDialogComponent implements OnInit {
+  protected readonly roles = AssetContactRoles.map((role) => ({
```
TIL-EBP, line 21: I can currently add the same contact with the same role, which does not really make sense. However, avoiding this might be a little tedious. in the base component you could check if the contact with that id and the same role already exists

TIL-EBP (approved):
It looks nice and most things work well. 
The only thing that is really breaking is that I cannot properly reset the form after having made changes to the contacts

libs/asset-editor/src/lib/components/asset-editor-tabs/asset-editor-general/asset-editor-general.component.ts
Tugark: Is it merged already? Because I'm pretty sure I rebased all the things?

libs/asset-editor/src/lib/components/asset-editor-tabs/asset-editor-contacts/asset-editor-contacts.component.ts
Tugark: No, you are absolutely right. That was also exactly a buggy behaviour. I changed it to remove the items in reverse order, because changing the form content directly is not recommended (and using `clear()` first and readding items leads to flickering).

libs/asset-editor/src/lib/components/asset-editor-tabs/asset-editor-general/asset-editor-general.component.ts
TIL-EBP: i might be tripping becuase it's not in my other open pr... maybe i just thought about doing it and then never did

libs/asset-editor/src/lib/components/asset-editor-tabs/asset-editor-contacts/link-contact-dialog/link-contact-dialog.component.ts
Tugark: I changed the logic, because with the current setup, we always add one contact id only, so we can take the id of the first element and remove them.

There is no error though, however, no duplicate entries are added anymore.

libs/asset-editor/src/lib/components/asset-editor-tabs/asset-editor-contacts/link-contact-dialog/link-contact-dialog.component.ts
Tugark: The flicker will still happen, because you close the dialog and open the next, and on one tick, you'd have no overlay. The problem is a bit because we have a second modal; so we'd have to refactor this into one modal and generically switch the forms. But that sounds like a lot of work, so I'm not sure if we should just merge it as is?

PR #539, opened 2025-05-08
Reduce /search response size

Tugark:
As discussed in today's daily, as a consequence of @TIL-EBP and my analysis of #475 , a "quick win" seemed the reduction of the response size of the `/search` endpoint. I did this in a timeboxed fashion, because we were quicker than expected with our concept in #475, so this seemed like a good opportunity to increase performance/page load time as well as refactoring some more `fp-ts` dependencies.

* Reduces the response size of the `/search` endpoint (on my database, reduces the swisstopo results filter from 2.7MB to 700kb)
* Refactors all related interfaces/classes from `shared` to `shared/v2` and getting rid of `fp-ts`
* Removed some unused classes (code cleanup)

I decided to open this PR into `feature/workflows` because otherwise, due to the recently performed linting, this could become a rather ugly merge/rebase when we are going to merge the workflows.

<hr>
Note: Further improvements would include reducing elasticsearch imports, because we don't need the full object in elasticsearch. However, that's an issue for another day :)


libs/shared/v2/src/lib/models/asset-search.ts
```diff
@@ -0,0 +1,54 @@
...
+}
+
+export interface AssetSearchResultItemContact {
+  role: AssetContactRole;
+  contactId: number;
+}
+
+export interface AssetSearchResultItem {
+  assetId: number;
+  createDate: number & DateIdBrand;
```
daniel-va, line 20: Can we replace this with either `Date` (for datetime) or `LocalDate` (exact date)?
I'm not the biggest fan of this `DateIdBrand` thing - this is a legacy type that we never got around to remove fully. Newer types normally don't use it anymore.

libs/shared/src/lib/models/asset-search/asset-search-result.ts
```diff
@@ -1,20 +1,8 @@
...
-  data: AssetEditDetail[];
-}
-
-export interface PageStats {
-  size: number;
-  offset: number;
-  total: number;
-}
-
 export interface AssetSearchStats {
```
daniel-va, line 6: Why do you not just move these types into `@shared/v2` as well why you're at it?

libs/asset-viewer/src/lib/services/asset-search.service.ts
```diff
@@ -1,26 +1,19 @@
...
 
   public search(searchQuery: AssetSearchQuery): Observable<AssetSearchResult> {
     return this._httpClient.post('/api/assets/search?limit=1000', searchQuery).pipe(
-      map((res) => plainToInstance(AssetSearchResultDTO, res)),
-      tap((result) => {
-        result.data = result.data.map((asset) => (AssetEditDetail.decode(asset) as E.Right<AssetEditDetail>).right);
+      map((res) => {
+        return plainToInstance(AssetSearchResultDTO, res);
       }),
     );
```
daniel-va, line 18: Very minor syntax change, would make this method align with the rest of this class.
```suggestion
      map((res) => plainToInstance(AssetSearchResultDTO, res)));
```

apps/server-asset-sg/src/features/assets/assets/search/asset-search.service.ts
```diff
@@ -690,6 +694,37 @@ export class AssetSearchService {
       }),
     };
   }
+
+  private parseSerializedAsset(serializedAsset: SerializedAssetEditDetail): AssetSearchResultItem {
```
daniel-va, line 698: Why can't we use a dto class based on [class-transfomer](https://www.npmjs.com/package/class-transformer) for this denormalization step?

daniel-va (approved):

libs/shared/src/lib/models/asset-search/asset-search-result.ts
Tugark: As discussed in teams: We wait for a proper refactoring, since moving this trickles down to basically all the types (e.g. via `UsageCode` and `LV95`).

I'll add a todo.

libs/shared/v2/src/lib/models/asset-search.ts
Tugark: As discussed - this would be optimal, but requires changing basically all the current workflows (up to and including `AssetEditDetail`). This should be changed when we refactor those parts.

This property is in a special format (`YYYYMMDD` as `number`); so we need another parse method. Furthermore, in order to keep payload size small, we should use to different transformers (`{toClassOnly: true}` / `{toPlainOnly: true}`) to send the number from the endpoint and only cast it to a proper date in the apps.

apps/server-asset-sg/src/features/assets/assets/search/asset-search.service.ts
Tugark: @daniel-va Done; I also changed our Controller to use the officially recommended middleware to return `class-transformer` objects (https://docs.nestjs.com/techniques/serialization, see notes on better typesafety).

libs/shared/v2/src/lib/models/asset-search.ts
```diff
@@ -33,22 +36,94 @@ export interface AssetSearchResult {
...
+  @IsString()
+  studyId!: string;
+
+  @Expose()
+  @IsString()
+  geomText!: string;
+}
+
+export class AssetSearchResultItemContactDTO implements AssetSearchResultItemContact {
+  @Expose()
```
daniel-va, line 50: I don't think you need these? We never had to use them up to now.

libs/shared/v2/src/lib/models/asset-search.ts
```diff
@@ -33,22 +36,94 @@ export interface AssetSearchResult {
...
+  publicUse!: AssetSearchResultItemUse;
+}
 
 export class AssetSearchResultDTO implements AssetSearchResult {
+  @Expose()
   page!: PageStats;
-  data!: AssetSearchResultItem[];
+  @Expose()
+  @Type(() => AssetSearchResultItemDTO)
+  data!: AssetSearchResultItemDTO[];
```
daniel-va, line 113: I personally like to type DTO fields by interface rather than by concrete class - IMO the DTO should really only be used for parsing, and the interface everywhere else.
```suggestion
  data!: AssetSearchResultItem[];
```

daniel-va (approved):

libs/shared/v2/src/lib/models/asset-search.ts
Tugark: This is now required due to how `{excludeExtraneousValues: true}` works; and this setting is now required to prevent all serialized values from being passed.

PR #551, opened 2025-05-19
Feature/assets 456 add page count to file

### LINKED ISSUE #456: Anzeige der Seitenanzahl von PDF-Dokumenten
**Beschreibung**

Die Anzahl der Seiten eines angehängten Dokuments soll im UI angezeigt werden.  Die Logik zur Ermittlung der Seitenanzahl soll im OCR-Repository implementiert werden, da das Dokument ohnehin zur Texterkennung dorthin gesendet wird. In der Asset-Suite sollen die notwendigen Anpassungen vorgenommen werden, um die Seitenanzahl anzuzeigen.

**Vorgeschlag zur Implementierung**
- Anzeige der extrahierten Seitenanzahl in der UI für Dokumente.
- Implementierung der notwendigen Logik zum Abrufen und Speichern der Seitenanzahl.
- Klärung, welche Daten genau benötigt werden vom OCR-Repository (z. B. API-Endpunkt, Datenformat).
- Die Seitenanzahl in der UI sollte auf Dokumentenebene angezeigt werden.

**Offene Fragen zur Abstimmung mit dem OCR-Repository**
- Wie soll die Asset-Suite die Seitenanzahl abrufen? Wird die Information gepusht oder muss sie abgefragt werden?
- In welchem Format soll die OCR-API die Seitenanzahl liefern? (z. B. als Teil der Metadaten oder über einen separaten Endpunkt?)
- Soll die Seitenanzahl dauerhaft in der Asset-Suite-Datenbank gespeichert oder nur dynamisch angezeigt werden?
- Wie wird mit Fehlerfällen umgegangen?

**Akzeptanzkriterien**

- [ ] Asset-Suite kann die Seitenanzahl eines Dokuments aus dem OCR-Repository abrufen.
- [ ] Die Seitenanzahl wird in der UI auf Dokumentenebene angezeigt.
- [ ] Die Implementierung erfolgt in Abstimmung mit der Implementierung im OCR-Repository, um die benötigten Daten korrekt zu erhalten.
- [ ]  Fehlerfälle werden sinnvoll behandelt (z. B. Standardwert oder Fehleranzeige).

Tugark:
Closes #456


apps/server-asset-sg/src/features/files/file-s3.service.ts
```diff
@@ -78,14 +71,22 @@ export class FileS3Service {
...
       return handleS3Error(e) ?? false;
     }
   }
 
+  private createMetadata(output: HeadObjectCommandOutput, name: string): FileS3Metadata {
+    return {
+      name,
+      byteCount: output.ContentLength ?? null,
+      mediaType: output.ContentType ?? null,
+      pageCount: Number(output.Metadata?.pagecount) ?? null,
```
daniel-va, line 87: What is the expected behavior with this null coalesce here? If it is meant to be set to `null` when the `pagecount` doesn't exist, then it doesn't work right now, as `Number` returns `NaN` for any non-number, which doesn't coalesce.
```suggestion
      pageCount: output.Metadata?.pagecount ? Number(output.Metadata.pagecount) : null,
```

development/docker-compose.yaml
```diff
@@ -13,6 +13,20 @@ services:
...
+      - storage
+    volumes:
+      - ./init/minio/:/init
+    entrypoint: >
+      /bin/sh -c "
+      /usr/bin/mc alias set storage http://storage:9000 ${STORAGE_USER} ${STORAGE_PASSWORD};
+      /usr/bin/mc mb --ignore-existing storage/asset-sg;
+      /usr/bin/mc admin accesskey create storage ${STORAGE_USER} --access-key ${STORAGE_ACCESS_KEY} --secret-key ${STORAGE_ACCESS_KEY_SECRET} --policy /init/policy.json;
+      exit 0;
+      "
```
daniel-va, line 28: Couldn't we simply move this init script into the MinIO container itself? I'm not the biggest fan of using temporary containers in compose when the script could just be run at the start of a persistent service. This also has the benefit of keeping the things that are relevant to MinIO within its service.

I don't think this is a blocking issue and am fine with leaving it as-is, I simply prefer not having it like this.

daniel-va (approved):

apps/server-asset-sg/src/features/files/file-s3.service.ts
Tugark: Good catch, that was my intention :)

development/docker-compose.yaml
Tugark: I absolutely agree with you, but as far as I know, `minio` image does not have an `init` option as postgres does. The way I did it is based on those issues that I was able to finde (https://github.com/minio/minio/issues/11604, https://github.com/minio/minio/issues/4769). We _could_ add the buckets, but I did not find a better solution. Let's leave it for now.

PR #554, opened 2025-05-19
Feature/assets 383 asset edit

### LINKED ISSUE #383: Umbau UI Bearbeitungsansicht Asset
**Beschreibung**

Die Bearbeitungsansicht eines Assets soll ebenfalls in die neue UI/UX Welt gebracht werden.

**Acceptance Criteria**

- [ ] Man kann ein Assets in der neuen Bearbeitungsansicht bearbeiten und speichern.

TIL-EBP:
resolves #383


libs/client-shared/src/lib/components/text-input/text-input.component.scss
```diff
@@ -1,5 +1,12 @@
 @use "../../styles/variables";
 
+::ng-deep .mat-mdc-option.author > .mdc-list-item__primary-text > .author__name {
```
Tugark, line 3: Not part of your PR, but why is this called `author__name` inside this generic text input?

libs/asset-editor/src/lib/components/asset-editor-tabs/asset-editor-references/add-reference-dialog/add-reference-dialog.component.ts
```diff
@@ -78,9 +79,7 @@ export class AddReferenceDialogComponent implements OnInit {
                 })
                 .pipe(
                   map((res) => plainToInstance(AssetSearchResultDTO, res)),
-                  map((result) =>
-                    result.data.map((asset) => (AssetEditDetail.decode(asset) as E.Right<AssetEditDetail>).right),
-                  ),
+                  map((res) => res.data),
```
Tugark, line 82: Combine this with the `map` above, e.g.:

```ts
.pipe(
              map((res) => {
                    const data = plainToInstance(AssetSearchResultDTO, res);
                    return data.data;
                  })
                )
````

File: libs/asset-editor/src/lib/components/asset-editor-page/asset-editor-page.component.ts [hunk 3]
```diff
@@ -185,17 +199,112 @@ export class AssetEditorPageComponent implements OnInit, OnDestroy {
...
+      .filter((file) => !filesToDelete.includes(file.value.id) && !file.value.file)
+      .map((file) => file.value);
+
+    const patchAsset: PatchAsset = {
+      titlePublic: general.titlePublic.value,
+      titleOriginal: general.titleOriginal.value,
+      createDate: dateIdFromDate(general.creationDate.value ?? new Date()),
+      receiptDate: dateIdFromDate(general.receiptDate.value ?? new Date()),
+      assetLanguages: general.assetLanguages.value,
+      workgroupId: general.workgroupId.value!,
```
Tugark, line 235: Are these non-null assetions due to some constraints with how they are getting set/reset? Or why can't we make those `nonNullable: true` as well?

libs/asset-editor/src/lib/components/asset-editor-page/asset-editor-page.component.ts
```diff
@@ -185,17 +199,112 @@ export class AssetEditorPageComponent implements OnInit, OnDestroy {
       });
   }
 
+  public save() {
```
Tugark, line 202: Personally, I'd move this logic to the `AssetEditoService`, since it is not directly tied to the presentation logic.

Also, I would move some bits of it into seperate functions (e.g. the whole `filesToDelete` things, because this method is really large as-is.

However, up to you, since some of the complexity comes from the fact that we did the whole form right in this component, which is not part of your PR.

Tugark (approved):

libs/client-shared/src/lib/components/text-input/text-input.component.scss
TIL-EBP: I asked myself the same thing I don't know tbh.

libs/asset-editor/src/lib/components/asset-editor-page/asset-editor-page.component.ts
TIL-EBP: If i make them non-nullable, I would also have to give them an initial value. And i am not sure this is what we want.

Implicitly, we know that the values are there becuase you can only reach this  method once the form is valid. 

So there are a couple of ways we can go here: 
- Set inital values for the required fields to make them nonNullable
- Always check if they are present (even though we know they are)
- Leave it as is

Thoughts?

libs/asset-editor/src/lib/components/asset-editor-page/asset-editor-page.component.ts
TIL-EBP: Will there be a lot of logic added to this with the workflows? if so, i would be in favor of moving it to the service, otherewise i would just leave it here

Tugark: I think there should be no logic in these parts, as the workflow stuff only targets `/workflows` endpoints, no?

The only thing that comes to mind is setting an asset's workflow status when saving, but that's not part of this PR as well.

libs/asset-editor/src/lib/components/asset-editor-page/asset-editor-page.component.ts
Tugark: Yeah, the "typesafe" solution would be to add a pre-defined value (e.g. `-1`) as default and create validators and template logic to handle this. But that would be overkill. And since it is only in this controlled part (and Angular forms are wonky anyways), it should be fine.

However - array controls like `manCatLabelRefs` _could_ be made `nonNullable`, right? Because they already do not allow `null` by setting a default value.

PR #556, opened 2025-05-20
DB Cleanup

### LINKED ISSUE #466: Datenbankschema bereinigen
Die meisten Tabellen/Spalten in der Postgres-Datenbank wurden 1-1 übernommen aus der alten InfoGeol-Datenbank. Dabei sind auch einige Tabellen/Spalten die nie wirklich benutzt werden (keine Daten, keine Verknüpfung mit irgendwelche Funktionalitäten in der Applikation), und das auch in Zukunft nicht geplant ist. Für diese Tabellen/Spalten soll verifiziert werden, dass sie wirklich nicht benutzt werden, und wenn ja, sollten sie gelöscht werden, damit das Datenbankschema einfacher und leichter zu verstehen wird.

**Composition**
Tabellen
- `asset_format_composition`
- `asset_kind_composition`

**Internal project**
Tabellen
- `internal_project`
- `asset_internal_project`

**Automatic categorization**
Tabellen
- `auto_cat`
- `auto_cat_label_item`
- `auto_object_cat_item`

Die letzte zwei Tabellen beinhalten Codelisten, die aber nicht mehr relevant sind.

**Asset object**
Tabelle
- `asset_object_info`

**Publication**
Tabellen
- `publication`
- `asset_publication`
- `pub_channel_item`

Die letzte Tabelle beinhaltet eine Codeliste, die aber nicht mehr relevant ist.

**Geometry**
Tabelle
- `geom_quality_item`

Beinhaltet eine Codeliste, die aber nicht mehr relevant ist.

Spalten
- `study_area.geom_quality_item_code`
- `study_location.geom_quality_item_code`
- `study_trace.geom_quality_item_code`

Im Moment werden hier nur die Werte "unkown" (sic, ohne "n" in der Mitte) und "revised" benutzt. Das könnte also auch einen einfachen Boolean Wert `is_revised` werden. Diese Werte sind (soweit ich sehen kann) in der Applikation nicht sichtbar / editierbar.

Vielleicht wäre es auch sinnvoll, die drei Tabellen `study_area`, `study_location` und `study_trace` zusammenzufügen. In der Applikation werden sie eigentlich schon so benutzt, über einen View `all_study`. Die neue Tabelle könnte `geometry` heissen ("Study" ist sowieso ein komischer Name) und diese Spalten haben:
- `geometry_id`
- `asset_id`
- `geom`
- `geometry_type` (mögliche Werte "point", "line", "polygon")

Tugark:
Resolves #466 

*General*
* Tables that are empty and not referenced in edit mode should be removed; if they become relevant, they should be readded.
* Item tables (`pub_channel_item`, `auto_cat_label_item`, `auto_object_cat_item`, `geom_quality_item`) did contain (unused) data, so they should be exported and stored somewhere for prod data (just in case :))

*Cleanup log for #466 mentioned tables*

* `asset_format_composition`: Empty on prod, int, dev; referenced in GUI (but always default) -> **dropped**
* `asset_kind_composition`: Empty on prod, int, dev; not referenced in GUI -> **dropped**
* `internal_project` & `asset_internal_project`: Empty on prod, int, dev; not referenced in GUI -> **dropped**
* `auto_cat` & `auto_cat_label_item` & `auto_object_cat_item` & `asset_object_info`: Empty on prod, int, dev (or, for _item tables, containing keys that are not used according to #466); not referenced in GUI -> **dropped**
* `publication` & `asset_publication` & `pub_channel_item`: empy on prod, int, dev (or, for _item tables, containing keys that are not used according to #466); not referenced in GUI -> **dropped**
* `geom_quality_item`: Another lookup table, which is used as a reference (via `geomQualityItemCode`) on `study_area`/`study_location`/`study_trace`. However, as mentioned in #466, only `revised` and `ukown` are currently present. The only other mentions I found is an import script (`/scripts/2024-10-01_import-new-geometries/main.py`) that sets them to revised; but there is no possibility to change these values. As such, I kept the value for consistency reasons (but see todos)
* Moving all `study_` tables into one table is not an optimal solution - because we have different geometry types, it can be worthwhile to have them seperate, also in order to create optimized where clauses, index lookups; and it is also easier to see what geometry a given record has (e.g. `st_geometrytype(geom)`.

*Additional cleanup steps*
* `study_location`/`study_trace`/`study_area`: These tables were using a very basic `geometry` type without a specified SRID. This has been changed so the columns have a proper geometry type and an SRID set.
* `all_study` View simplified - there is much more potential in optimizing this, but changing stuff (e.g. removing `study_` prefix) has a lot of consequences (e.g. handling proper table for inserting geometries) -> placed in #553 

*To check*
- [ ] Should the `geomQualityItemCode` on `study_*` tables be removed? Or transformed to boolean?
- [ ] I also changed `seed-assets` and stuff; but they are probably no longer used. Revert, or keep, if at some point we'd need to seed?


daniel-va (approved):

daniel-va:
- IMO we can just drop `geom_quality_item_code` on the study tables, I don't think we use them anywhere. I would say if @stijnvermeeren-swisstopo thinks we can get rid of them, then I would prefer not having them at all. On the same note, I would actually also remove these columns in all the other tables. I understand your reasoning why you kept them, but the script that references them is quite old (not used by us), and if no one at swisstopo needs this data, then I don't see a reason to keep it.

- `seed-assets` is legacy from before we took over this project, and we never bothered to adjust it. We could actually get rid of it in the planned cleanup sprint.

daniel-va (approved):

Tugark:
As discussed with @stijnvermeeren-swisstopo and  @mPfifi , `is_revised` added as a new "backup" column.

PR #582, opened 2025-06-03
Feature/assets 499 sync service

### LINKED ISSUE #499: Update sync service
**Beschreibung**

Folgend #492, passe den Sync Service so an, dass:
- Nur Assets, deren Workflow status `reviewed` und `isPublished=true` ist, werden publiziert.
- Für Assets, die publiziert werden, werden nur Informationen publiziert, für deren Bereich sowohl bei Review als auch Freigabe ein Haken gesetzt ist.

**Acceptance Criteria**

- [ ] Der Sync Service folgt den neuen Publikationsregeln mithilfe Workflows

Tugark:
Closes #499 

* Removes all the old workflow elements (tables, attributes)
* Removes controller authorization logic related to these old workflow elements (see #581 , this should all be done there)
* Simplifies usage to a boolean flag (while still using `makeUsageCode` for simplicity reasons)
* Creates data migrations according to the specification in #499
* Updates the SyncService for view (and _only_ for view, not for extern!) to have the new publication logic
* Removed `processor` and `lastUpdatedAt` columns and references - if this is required, we'll have to re-add it later on (mainly in EditDetail view).
* Added todos for @TIL-EBP where the `isPublic` handling needs to be implemented

To check, I exported the prod database and have the following counts:
* `asset`: View Prod = 1070, View Local = 1070
* `study_area`: View Prod = 632, View Local = 632
* `study_location`: View Prod = 4113, View Local = 4113
* `study_trace`: View Prod = 54, View Local = 54
* `asset_x_asset_y`: View Prod = 322, View Local = 322
* `asset_file`: View Prod = 1059, View Local = 1059
* `file`: View Prod = 1054, View Local = 1054
* `contact`: View Prod = 0, View Local = 0
* `asset_contact`: View Prod = 0, View Local = 0

I also tried setting the values manually and contacts and legal files are synced correctly.

Note that we don't have any exported contacts on PROD (!) and, of course, due to our migration, none on  local (because they should not be exported right now).

Last, but not least: There was a flaky test - but I cannot reproduce it and it does no longer fail. Just in case we get failures down the road :)


apps/sync-asset-sg/.env
```diff
@@ -13,4 +13,8 @@ DESTINATION_S3_BUCKET_NAME=asset-sg
 DESTINATION_S3_ASSET_FOLDER=asset-sg
 
 # Allowed workgroup names (comma-separated)
-ALLOWED_WORKGROUP_IDS='1'
+SOURCE_WORKGROUP_IDS='1'
+DESTINATION_WORKGROUP_IDS='1'
+
+# view|extern
+MODE='view'
```
daniel-va, line 20: Not sure we want to merge these - are these not just here because you needed to run the sync locally?

apps/sync-asset-sg/src/core/export-to-view.service.ts
```diff
@@ -182,38 +134,78 @@ export class ExportToViewService {
...
+   * from other elements, i.e. numbers. For safety, these are still handled via parameters (note: SQL params start at 1,
+   * like $1, $2, ...)
+   */
+  private async exportGeometries(assetIds: number[], table: 'study_area' | 'study_location' | 'study_trace') {
+    const query = `
+      SELECT ${table}_id as "id",
+             asset_id    as "assetId",
+             st_astext(geom, 2056) as geom
+      FROM ${table}
+      WHERE asset_id IN (${assetIds.map((_, idx) => `$${++idx}`).join(',')})
```
daniel-va, line 186: AFAIK Postgres has a limit on variables in a prepared statement at around 65'000, meaning that once we have 100'000+ assets, this could fail - maybe we should batch the geometries export too?

apps/sync-asset-sg/src/core/log.ts
```diff
@@ -1,3 +1,3 @@
-export function log(log: string) {
-  console.log(`[${new Date().toISOString()}] ${log}`);
+export function log(log: string, level: 'main' | 'batch' = 'batch') {
```
daniel-va, line 1: Not sure about using `batch` as the default here - you use that default exactly once, while `main` is used a lot more.

libs/persistence/prisma/scripts/seed-assets.ts
```diff
@@ -136,8 +136,6 @@ export const importAssets = async () => {
     municipality: parsed[5] || null,
```
daniel-va, line 136: We could just remove this file, since it is more confusing than anything else...

daniel-va (approved):

apps/sync-asset-sg/.env
Tugark: As discussed - this is needed because they are checked in `main.ts` and throw if not found.

libs/persistence/prisma/scripts/seed-assets.ts
Tugark: Removed the whole folder since it is no longer usable.

apps/sync-asset-sg/src/core/export-to-view.service.ts
Tugark: Good catch, added a configurable batch size for geometries as well (y).

PR #589, opened 2025-06-10
Asset #554: Fix translations in Edit UI

TIL-EBP:
resolves #554

### DIFF
--- a/libs/client-shared/src/lib/components/select/select.component.html
+++ b/libs/client-shared/src/lib/components/select/select.component.html
@@ -21,7 +21,7 @@
       }
 
       @for (option of customOptionsToRender; track option.value) {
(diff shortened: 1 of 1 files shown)


libs/client-shared/src/lib/components/select/select.component.html
```diff
@@ -15,7 +15,7 @@
         </mat-select-trigger>
       }
       @for (value of values; track value) {
-        <mat-option [value]="value">
+        <mat-option [value]="value" class="option">
```
daniel-va, line 18: This class should probably also be added on line 24. Or, optionally, why do you need a class to style this? You could just style every mat-option, regardless of class.

daniel-va (approved):

PR #616, opened 2025-06-23
Elastic Search

### LINKED ISSUE #580: ElasticSearch Sync bei Deployment/Neustart automatisieren
**Beschreibung**

Eine klare und präzise Beschreibung der Ausgangssituation, der Problematik sowie bereits bekannter Lösungsansätze.

**Abgrenzung**

Nicht beachtet in dieser Aufgabe wird die Suche nach der Frage.

**Acceptance Criteria**

- [ ] Die Antwort muss 42 sein.
- [ ] ...

Tugark:
Closes #580 

Implemented as restart hook; will trigger on boot.

Reasoning: See comment history, TLDR: k8s approach with init containers has too much overhead.

Drawbacks: _if_ we scale out, this can of course lead to race conditions, but this would be another issue anyway (i.e. if multiple users sync on different shards).


daniel-va (commented):
Seeing the large block of boilerplate in `deployment.api.yaml`, I think it would actually make more sense to just integrate this behavior directly into the API. To me, it seems like we could then replace all of the changes in here with a single additional line of code.

daniel-va (approved):

PR #640, opened 2025-07-09
Fix broken filters

Tugark:
Fixes the bugs @mPfifi reported in teams (missing filters for topics; non working search terms).

* Missing filters: There was an issue with the rename from `manCatLabels` to `topicCodes`, as well as an issue with items now being parsed as `Map` (which cannot be used directly with `Object.values()`
* Non working search terms: This issue arose because in #579 , `escapeElasticQuery` was reintroduced (which was only present in the deprecated and now-removed `searchOld`): while technically correct, this also prevents @mPfifi's search term (`*tech*`) due to escaped characters. See  #639 for a follow up; this needs further discussion.


vej-ananas (approved):
marvelous work.

PR #660, opened 2025-07-21
Assets 653: Freigabe Haken

TIL-EBP:
Approval checkboxes are unchecked when review checkboxes are modified after staus has changed back to review. 
Required changes to core lib: https://github.com/swisstopo/swissgeol-ui-core/pull/74

### DIFF
--- a/libs/asset-editor/src/lib/components/asset-editor-tabs/asset-editor-status/asset-editor-status.component.ts
+++ b/libs/asset-editor/src/lib/components/asset-editor-tabs/asset-editor-status/asset-editor-status.component.ts
@@ -15,7 +15,7 @@ import { TranslateService } from '@ngx-translate/core';
 import { type SgcWorkflowSelectionEntry, WorkflowChange } from '@swissgeol/ui-core';
 import { SgcWorkflowChangeEvent } from '@swissgeol/ui-core/dist/types/components/sgc-workflow/sgc-workflow';
 import { SgcWorkflowSelectionChangeEvent } from '@swissgeol/ui-core/dist/types/components/sgc-workflow/sgc-workflow-selection/sgc-workflow-selection';
-import { BehaviorSubject, switchMap } from 'rxjs';
+import { BehaviorSubject } from 'rxjs';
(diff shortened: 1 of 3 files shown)


daniel-va (approved):

libs/asset-editor/src/lib/components/asset-editor-tabs/asset-editor-status/asset-editor-status.component.ts
```diff
@@ -70,19 +70,37 @@ export class AssetEditorStatusComponent implements OnChanges {
...
-          workflow: {
-            ...workflow,
-            review: {
-              ...workflow.approval,
-              ...patch,
+    this.workflowApiService
+      .updateReview(workflow.id, patch)
+      .pipe(
+        switchMap(() =>
+          this.workflowApiService.updateApproval(
```
daniel-va, line 77: Just for clarity, this resets the approval for fields whose review state has changed, right? Why do we do this via client request, and not automatically in the API?

TIL-EBP: I initially just modified the method on the backend that sets  the review fields to also reset the approval fields.
But then I thought its clearer if we keep both enpoints completely separate. 
Also I am using the return vallue from the approval endpoint to update the checkboxes correctly and I found it weird to return the approval data from the review enpoint

TIL-EBP: II do agree that this would reduce the risk of forgetting if we ever trigger this from another place. What would you then return from the review enpoint? Approval and Review? the whole workflow? Just returning the approval from the review endpoint seems wrong to me.

Or would you not use the return value at all and update the checkboxes in the frontend directly?

daniel-va: In this situation, I would not use the return value and instead update the approval manually. Keeps the API clean and shouldn't be too much additional code, no?

PR #684, opened 2025-08-25
Add extraction API

### LINKED ISSUE #675: Extraction API in Assets einbinden
**Beschreibung**

Analog zum OCR Service soll die neue API beim Upload von Files aufgerufen werden
- Erst nach erfolgreichem OCR-Durchlauf

Datenbankanpassungen: 
- Neue Tabelle `FilePage`:
  ```ts
	enum FilePageCategory {
	Text,
	Boreprofile,
	Maps,
	TitlePage,
	}

	FilePage {
	  fileId: FileId
	  pagenumber: int
	  categories: FilePageCategory[]

	}
  ```
Anpassung im `OcrStatus` und `File`:

```ts
  enum FileProcessingStage {
  Ocr,
  Extraction,
  Done
  }

  // Replaces OcrStatus, removes 'created'
  enum FileProcessingState {
    WillNotBeProcessed
    Waiting
    Processing
    Error
    Success
}

File {
...
processingStage: FileProcessingStage,
processingState: FileProcessingState
}

```

**Acceptance Criteria**

- [ ] Nach dem OCR Service wird der Extraction Service aufgerufen und die Stage und States korrekt angepasst.
- [ ] Der Service ist getestet

Tugark:
Closes #675 

* Refactored processing services to use events for dispatching (yes, events and signals _can_ be the root of all evil because they hide dependencies, but in this case, it is very isolated, contentrated and limited, so it is fine; and, as a bonus, we can easily move to proper worker queues (which would be better anyway)
* Added docker compose setup for the API - NOTE: this currently requires a manual build of the extraction api.

Open issues:
* Currently, API returns an array of extraction results, is there a reason?
* Not sure if all helm charts are properly updated
* How do we handle the deployment? I could add a feature flag with an env variable or another release branch?

DO NOT MERGE

### DIFF
--- a/libs/shared/v2/src/lib/models/asset-file.ts
+++ b/libs/shared/v2/src/lib/models/asset-file.ts
@@ -38,7 +38,7 @@ export interface AssetFile {
   lastModifiedAt: Date;
 
   /**
-   * DB: `file_processing_state`
+   * DB: `file_processing_stage`
    */
   fileProcessingStage: FileProcessingStage | null;
 


apps/server-asset-sg/src/features/assets/files/file-processors/file-ocr.service.ts
```diff
@@ -0,0 +1,61 @@
...
+
+  protected async postProcess(file: ProcessableFile, _: []): Promise<void> {
+    const data: Prisma.FileUpdateInput = {};
+
+    const metadata = await this.fileS3Service.loadMetadata(file.name);
+    if (metadata === null) {
+      this.logger.error('Processed file not found in S3', { file: file.name });
+      data.fileProcessingState = FileProcessingState.Error;
+    } else {
+      data.size = metadata.byteCount ?? 0;
```
daniel-va, line 47: Do we really want to set the file size to `0` when we don't know how large the file is? Doesn't this have the potential to overwrite correct information?

libs/shared/v2/src/lib/models/asset-file.ts
```diff
@@ -38,9 +38,36 @@ export interface AssetFile {
   lastModifiedAt: Date;
 
   /**
-   * DB: `ocr_status`
+   * DB: `file_processing_state`
```
daniel-va, line 41:
```suggestion
   * DB: `file_processing_stage`
```

daniel-va (approved):

apps/server-asset-sg/src/features/assets/files/file-processors/file-ocr.service.ts
Tugark: The information is no longer correct when we are here, but the interface allows `byteCount` to be null. So in the worst case, we have stored the "original" filesize, which is no longer valid after OCR because the file size changes.

PR #701, opened 2025-09-09
Set workflow.status to "Draft" if it does not exist (relevant for viewer)

TIL-EBP:


Tugark (commented):
Can you add a comment and reference the ticket here: https://github.com/swisstopo/swissgeol-assets-suite/issues/702

PR #711, opened 2025-09-15
Feature 677: Manuelles Bearbeiten der Seiten-Klassifizierung - Testing

daniel-va:
Part of #677.

Re-initializes the cypress e2e tests so we are able to correctly test the new classification features.


apps/server-asset-sg/src/run-command.ts
```diff
@@ -0,0 +1,26 @@
+import 'reflect-metadata';
+
+import { exit } from 'process';
+import { CommandFactory } from 'nest-commander';
+import { AppLogger } from '@/app.logger';
+import { AppModule } from '@/app.module';
+
+async function bootstrap() {
```
Tugark, line 8: As discussed today, we should add an `oh snap` early return if we - by whatever ungodly accidents - were to run this in dev, int or prod environments.

Tugark (approved):
I glanced over it, but did not go into detail. Also, it's a bit a mess because I think the branches diverged since I forced push your origin branch, I can look over it in more depth if needed if we sort that out.

But in general, I think we can merge it, and as discussed today, it is a very good starting point for extending it in the future.

PR #775, opened 2025-11-17
Hotfix: Build Sync Service

daniel-va:


TIL-EBP (approved):

apps/sync-asset-sg/tsconfig.app.json
```diff
@@ -3,7 +3,14 @@
   "compilerOptions": {
     "outDir": "../../dist/out-tsc",
     "module": "commonjs",
-    "types": ["node"]
+    "types": ["node"],
```
Tugark: Why do we need them explicitly here? They are in the baseconfig either way, no?

Tugark: Also, this module now targets `es2021`, whereas our baseconfig still has `es2015`

package.json
```diff
@@ -158,6 +158,7 @@
     "postcss": "^8.4.5",
     "prettier": "~3.6.2",
     "prettier-plugin-prisma": "^5.0.0",
+    "shx": "^0.4.0",
```
Tugark: Why is this dependency needed? I may have overlooked it, but I don't see where it is being used?

PR #875, opened 2026-02-18
Load files from S3 and store their page fulltext content into the db

### LINKED ISSUE #859: Volltextsuche: Bestehende Texte laden
**Beschreibung**
Die bestehenden Texte, die bereits via S3 verfügbar sind, müssen in die DB gespeichert werden.

**Abgrenzung**
Die Indizierung in Elasticsearch sowie die Erweiterung der Suche werden nicht in diesem Ticket, sondern in #854 implementiert.

**Acceptance Criteria**

- [ ] Die Texte/Inhalte von Files sind in der DB verfügbar.

vej-ananas:
Resolves #859


apps/server-asset-sg/src/features/assets/sync/atomic-progress.service.ts
```diff
@@ -0,0 +1,73 @@
+import fs, { rm } from 'fs/promises';
+import { Injectable, Logger } from '@nestjs/common';
+
+@Injectable()
+export abstract class AtomicProgressService<T> {
```
Tugark: I would at least narrow it to `T extends object` since that's what you pass.

The cleaner solution would be to have an empty base  interface (which is sketchy in TS, I know) like `AtomicProgressOptions`, which is used in the generic (also as default) and then extended and passed in extending classes. But since empty interfaces are problematic, narrowing it to `extends object` is fine.

PR #879, opened 2026-02-27
Feature/assets 874 dedup files

vej-ananas:
prepares #874 
This includes the script which turns `many assets - assetFile - the same file` into  `many assets - assetFile - many files` so that the following transition to many-one can be done without losing info.

I am unsure if a Command in the Server is the proper place for this, since we only really run it once and then never again. Maybe you have suggestions on where to put this? Or do we not keep it at all?


Tugark (approved):
I think it's fine using a command, since we need service and a fully configured nestjs instance due to prisma and S3. Just make a mental note to remove it once the migration is deployed.

PR #889, opened 2026-03-05
Turn asset-file into one-to-many

vej-ananas:
relates to #874 

Only merge once the dedup is done!

### DIFF
--- a/apps/server-asset-sg/src/features/assets/files/file.repo.ts
+++ b/apps/server-asset-sg/src/features/assets/files/file.repo.ts
@@ -1,12 +1,4 @@
-import {
-  AssetFile,
-  AssetFileId,
-  AssetId,
-  FileProcessingStage,
-  FileProcessingState,
-  LegalDocCode,
-  User,
-} from '@asset-sg/shared/v2';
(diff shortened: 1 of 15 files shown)


apps/server-asset-sg/src/features/assets/files/file.repo.ts
```diff
@@ -62,11 +62,9 @@ export class FileRepo implements Repo<AssetFile, FileIdentifier, CreateFileData,
...
   async findOrphans(): Promise<AssetFile[]> {
-    const entries = await this.prisma.file.findMany({
-      where: { AssetFile: { none: {} } },
-      select: assetFileSelection,
-    });
-    return entries.map(mapAssetFileFromPrisma);
+    // With the one-to-many relationship (assetId is NOT NULL on file),
+    // orphaned files should not exist. This method is kept for safety.
+    return [];
   }
```
Tugark: Why for safety? We should be able to find any usages of it, and if there are any, we should remove them anyway?

Tugark (commented):
Looks good, but I added some questions for clarification.

Note that I did not check it out, as I'm currently in the mid of fixing another issue. Let me know if you think I should check it out and run it properly.

wendlans (approved):
LGTM as far as I can tell.

PR #893, opened 2026-03-11
Allow releases from branches named release-candidate/...

vej-ananas:
Allow releases from branches named release-candidate.


Tugark (approved):
Image

PR #894, opened 2026-03-11
Feature/assets 722 auth inconsistencies

### LINKED ISSUE #722: Bug: Inkonsistenzen Authentisierung
**Beschreibung**
Im Rahmen der Workgroups wurden verschiedene Aspekte überarbeitet. Uns sind zwei Inkonsistenzen aufgefallen:

1. Wenn ein User sich in die App einloggt, aber noch _keiner_ Workgroup zugewiesen ist, so werden ihm dennoch alle Punktgeometrien auf der Übersicht (inkl. Heatmap) angezeigt. Die Authorisierung greift erst, wenn der Nutzer versucht, ein Asset zu öffnen; dann erhält er einen Zugriffsfehler.
2. Wenn ein User "Administrator" ist, dann hat er automatisch lesenden Zugriff auf ALLE Workgroups, und kann diese Filtern und anschauen.

Es benötigt einen Testuser, dieser kann entsprechend manipuliert werden.

@mPfifi Sollen wir das bei Gelegenheit mal besprechen und anschauen?

Tugark:
Closes #722 

* Filters geometries by a user's workgroup as well to prevent all datapoints from being visible
* Admins now see all available workgroups in filters, with only those being available for selection that they have the access rights for; however, the actual amount of assets per group is still visible (as request by @mPfifi )

### DIFF
--- a/apps/server-asset-sg/src/features/assets/search/asset-search.controller.ts
+++ b/apps/server-asset-sg/src/features/assets/search/asset-search.controller.ts
@@ -55,15 +55,29 @@ export class AssetSearchController {
     query: AssetSearchQuery,
     @CurrentUser() user: User,
   ): Promise<AssetSearchStats> {
+    if (user.isAdmin) {
+      // For admins: get restricted stats (correct totals and filter counts),
+      // then overlay unrestricted workgroup counts for workgroup discovery.
+      const restrictedQuery = { ...query };
+      restrictQueryForUser(restrictedQuery, user);
+      const [restrictedStats, unrestrictedStats] = await Promise.all([
+        this.assetSearchService.aggregate(restrictedQuery, user),
+        this.assetSearchService.aggregate(query, user),
+      ]);
+      return {
+        ...restrictedStats,
+        workgroupIds: unrestrictedStats.workgroupIds,
+      };
+    }
     restrictQueryForUser(query, user);
     return await this.assetSearchService.aggregate(query, user);
   }
 }
 
+/**
+ * Restricts the search query to only include workgroups the user has explicit membership in.
+ */
 export const restrictQueryForUser = (query: AssetSearchQuery, user: User) => {
-  if (user.isAdmin) {
-    return;
-  }
   query.workgroupIds =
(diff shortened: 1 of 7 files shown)


apps/server-asset-sg/src/features/assets/search/asset-search.controller.ts
```diff
@@ -55,15 +55,29 @@ export class AssetSearchController {
     query: AssetSearchQuery,
     @CurrentUser() user: User,
   ): Promise<AssetSearchStats> {
+    if (user.isAdmin) {
+      // For admins: get restricted stats (correct totals and filter counts),
+      // then overlay unrestricted workgroup counts for workgroup discovery.
+      const restrictedQuery = { ...query };
+      restrictQueryForUser(restrictedQuery, user);
+      const [restrictedStats, unrestrictedStats] = await Promise.all([
+        this.assetSearchService.aggregate(restrictedQuery, user),
```
vej-ananas: This effectively doubles the aggregation for each admin request. Is this relevant? Would it be worth it to refactor it to be more efficient (e.g. exclude the workgroup filter for the workgroup stats query inside `assetSearchService.aggregate(..)`)?

vej-ananas (approved):

apps/server-asset-sg/src/features/assets/search/asset-search.controller.ts
Tugark: @vej-ananas No, you're right. While it might not be a huge performance cost, it's still not efficient. I just pushed a change that allows for an override of this property, potentially also allowing for more overrides if need be.

Could you have another look at it and tell me if you like this better?

apps/server-asset-sg/src/features/assets/search/asset-search.service.ts
vej-ananas: Remove duplicated comments

vej-ananas (changes requested):

apps/server-asset-sg/src/features/assets/search/asset-search.controller.ts
vej-ananas: @Tugark Yea thats much nicer 🙂

vej-ananas (approved):

PR #920, opened 2026-04-01
Classify vulnerabilities and schedule for fix

vej-ananas:


wendlans (approved):
`We will update the package soon, but for now it should be safe to ignore.`
Approved on that basis. Especially the _will update the package soon_ ™

PR #921, opened 2026-04-01
Cleanup command for file deduplication

vej-ananas:
#874
Cleanup command for file deduplication. It was applied to all environments and is no longer needed.


wendlans (commented):
@vej-ananas  Just for my understanding: This de-duplication is something that was done once and never again? So a file duplication will never happen again in the future?

vej-ananas:
> @vej-ananas Just for my understanding: This de-duplication is something that was done once and never again? So a file duplication will never happen again in the future?

Exactly! We change the database to a 1 to many in a following PR, so duplicates will not be possible moving forward.

wendlans (approved):

PR #924, opened 2026-04-16
Feature/assets 854 file search

### LINKED ISSUE #790: Volltextsuche: Sammelissue
**Beschreibung**

Neben Elastic Search für die Metadaten soll eine Volltextsuche in den pdf-Dateien gemäss den Vorgaben in Figma entwickelt werden (vgl. https://www.figma.com/design/cEiOoOazAQZqpRY92ZhBeO/swissgeol?node-id=14124-241293&p=f).

Fortsetzung von Issue #25 
Dieses Issue wird aufgeteilt in mehrere Unteraufgaben, da es ein sehr Umfangreiches Feature ist.
Folgende Aufgaben stehen an: 

- Konzept (#853):  erarbeiten, wie der Text der Dateien von S3 in das Assets Backend gespiesen wird und in welcher Form er da gespeichert wird. In Absprache/ unter Berücksichtigung von #733 
- Backend (#854): Backend ergänzen um entsprechende Felder/Funktionalität, dass Volltexte gespeichert werden können und ElasticSearch um den Volltext Index ergänzt werden kann
- Frontend (#856): Umsetzung gemäss Figma und Spezifikation 

**Acceptance Criteria**

- [ ] Die pdf-Dateien/Dokumente lassen sich mittels Volltextsuche inhaltlich durchsuchen.

vej-ananas:
Resolves #790 #854 #856


.github/workflows/publish-prod.yml
```diff
@@ -55,7 +55,7 @@ jobs:
         run: |
           content=$(awk '
             BEGIN { found = 0 }
-            /^## \[/ {
+            /^## / {
```
wendlans, line 58: This is the only change in the workflows: was it done intentionally?

apps/client-asset-sg/src/app/i18n/translations-guidelines.md
```diff
@@ -9,3 +9,7 @@ Newly added Keys (since 22.01.2026)
 - pdfNoLongerAvailable
 - directLink
 - directLinkCopied
+
```
wendlans, line 12: What about `textMatch`, `page`, ...?

apps/server-asset-sg/src/features/assets/search/file-search.service.ts
```diff
@@ -0,0 +1,98 @@
...
+          },
+        },
+        countAggs: { totalAssets: { field: 'assetId' } },
+      },
+    );
+
+    const data: FileSearchResultItem[] = [];
+    for (const hit of results.values()) {
+      data.push({
+        fileId: Number(hit.source?.['fileId']),
```
wendlans, line 67: Just a quick question: It's not a problem that this can become `NaN`, right? (same with the other properties)

apps/server-asset-sg/src/features/assets/search/file-search-writer.service.ts
```diff
@@ -0,0 +1,369 @@
...
+  constructor(
+    private readonly elastic: ElasticsearchClient,
+    private readonly prisma: PrismaClient,
+    private readonly geometryRepo: GeometryRepo,
+    private readonly options: SearchWriterOptions,
+  ) {
+    this.eager = options.isEager ? this.fetchEager() : Promise.resolve(null);
+  }
+
+  async write(oneOrMore: AssetFileId | AssetFileId[]): Promise<void> {
```
wendlans, line 56: Why aren't you just using `assetFileIds: AssetFileId[]`? If it's only one it's easy to just add it to an array before running this command.

apps/server-asset-sg/src/features/assets/search/file-search.writer.ts
```diff
@@ -0,0 +1,369 @@
...
+      index: this.options.index,
+      query: { terms: { fileId: fileIds } },
+      refresh: false,
+      ignore_unavailable: true,
+    });
+
+    const operations: Array<BulkOperationContainer | ElasticsearchFilePage> = [];
+
+    for (const file of fileRecords) {
+      const pages = file.fulltextContent as unknown as FulltextContent[] | null;
```
wendlans: I don't like this solution of using a double assertion. It would be better to use a helper function to check for the type and transform it properly.

apps/server-asset-sg/src/features/assets/search/file-search.writer.ts
```diff
@@ -0,0 +1,369 @@
...
+      ignore_unavailable: true,
+    });
+
+    const allOperations: Array<BulkOperationContainer | ElasticsearchFilePage> = [];
+
+    const processQueue = new ProcessQueue(QUEUE_SIZE);
+    const operationsByAsset: Array<Array<BulkOperationContainer | ElasticsearchFilePage>> = Array(assets.length);
+
+    for (let j = 0; j < assets.length; j++) {
+      const i = j;
```
wendlans: Why reassigning `j`?

apps/server-asset-sg/src/features/assets/search/file-search.writer.ts
```diff
@@ -0,0 +1,369 @@
...
+      const asset = assets[i];
+      processQueue
+        .add(async () => {
+          operationsByAsset[i] = await this.mapAssetFilesToElastic(asset);
+        })
+        .then();
```
wendlans: You could just use `await` at the beginning instead of `then()` here

apps/server-asset-sg/src/features/assets/search/file-search.writer.ts
```diff
@@ -0,0 +1,369 @@
...
+        fulltextContent: true,
+      },
+    });
+
+    return files
+      .filter((f) => f.fulltextContent != null)
+      .map((f) => ({
+        fileId: f.id,
+        fileName: f.nameAlias ?? f.name,
+        pages: f.fulltextContent as unknown as FulltextContent[],
```
wendlans: Same here: I don't like this solution of using a double assertion. It would be better to use a helper function to check for the type and transform it properly.

apps/server-asset-sg/src/features/assets/search/file-search.writer.ts
```diff
@@ -0,0 +1,369 @@
...
+        nameAlias: true,
+        fulltextContent: true,
+      },
+    });
+    const mapping = new Map<number, { fileName: string; pages: FulltextContent[] }>();
+    for (const file of files) {
+      if (file.fulltextContent != null) {
+        mapping.set(file.id, {
+          fileName: file.nameAlias ?? file.name,
+          pages: file.fulltextContent as unknown as FulltextContent[],
```
wendlans: And same here

apps/server-asset-sg/src/features/assets/search/file-search-writer.service.ts
wendlans, line 1: General question: when do you create unit tests in this project? This writer contains a lot of logic but not a single unit tests.
**EDIT**: same question for all other `util`, `service`, `writer` and other classes in the backend that contain a lot of logic.

apps/server-asset-sg/src/features/assets/search/search-query.utils.ts
```diff
@@ -0,0 +1,250 @@
+import { ElasticsearchAsset, SearchQueries, SearchType, User } from '@asset-sg/shared/v2';
+import {
+  AggregationsAggregationContainer,
+  QueryDslDateRangeQuery,
+  QueryDslQueryContainer,
+} from '@elastic/elasticsearch/lib/api/types';
+import { AGGREGATION_NUMBER_OF_BUCKETS, SEARCHABLE_FIELDS } from '@/features/assets/search/asset-search.constants';
+import { mapLv95ToElastic } from '@/features/assets/search/asset-search.utils';
+
+export interface PageOptions {
```
wendlans, line 10: Just a question: how do you organize files in this project? I'm seeing a mixture between functions, static methods, interfaces, normal classes all mixed in some files. Is there a structure to this?

libs/shared/v2/src/lib/models/search/search-query.ts
```diff
@@ -5,8 +5,17 @@ import { GeometryType } from '../geometry';
 import { LocalizedItemCode } from '../localized-item';
 import { WorkgroupId } from '../workgroup';
 
-export interface AssetSearchQuery {
+export enum SearchType {
```
wendlans, line 9: Prefer union types over enums in typescript.

apps/server-asset-sg/src/features/assets/search/search-query.utils.ts
```diff
@@ -0,0 +1,250 @@
...
+  limit?: number;
+  offset?: number;
+}
+
+export const mapQueryToElasticDsl = (query: SearchQueries, user: User): QueryDslQueryContainer => {
```
wendlans, line 15: Optional cosmetic: I'd prefer `function` instead of  `xyz = () => {}`
**EDIT**: I've just seen that this format is used everywhere in this file - you can ignore this comment.

apps/server-asset-sg/src/features/assets/search/search-query.utils.ts
```diff
@@ -0,0 +1,250 @@
...
+  if (query.authorId != null) {
+    filters.push({
+      term: {
+        authorIds: query.authorId,
+      },
+    });
+  }
+  if (query.createdAt != null && Object.keys(query.createdAt).length > 0) {
+    const createdAtFilter: QueryDslDateRangeQuery = {
+      format: 'yyyy-MM-dd',
```
wendlans: Magic string

apps/server-asset-sg/src/features/assets/search/search-query.utils.ts
```diff
@@ -0,0 +1,250 @@
...
+ */
+export const normalizeFieldQuery = (query: string): string =>
+  query
+    .replace(/title(_*)public:/gi, 'title:')
+    .replace(/title(_*)original:/gi, 'originalTitle:')
+    .replace(/contact(_*)name:/gi, 'contactNames:')
+    .replace(/asset(_*)id:/gi, 'id:')
+    .replace(/sgs(_*)id:/gi, 'sgsId:');
+
+export type MakeAggregationFunction = (
```
wendlans: You don't have to export this one. It's only used here. In exactly one place.

apps/server-asset-sg/src/features/assets/search/search-query.utils.ts
```diff
@@ -0,0 +1,250 @@
...
+ * Each aggregation bucket excludes its own field from the query filter so that
+ * selecting a value in one facet doesn't collapse that facet's own counts.
+ *
+ * When the caller passes a `queryOverride` (e.g. for the unrestricted workgroup bucket),
+ * that override is used as the base query instead of the default.
+ */
+export const createMakeAggregation = (query: SearchQueries, user: User): MakeAggregationFunction => {
+  return (operator, groupName, fieldName, queryOverride) => {
+    const baseQuery = queryOverride ?? query;
+    const { filter, aggs } = mapQueryToElasticDslParts({ ...baseQuery, [groupName]: undefined } as SearchQueries, user);
```
wendlans: What's going on here? why is the assertion needed?

apps/server-asset-sg/src/features/assets/search/search-writer.service.ts
```diff
@@ -0,0 +1,156 @@
...
+    await this.elastic.indices.create({ index: SYNC_INDEX });
+    await this.elastic.indices.putMapping({ index: SYNC_INDEX, ...indexMapping });
+
+    await this.elastic.indices.create({ index: FILE_SYNC_INDEX });
+    await this.elastic.indices.putMapping({ index: FILE_SYNC_INDEX, ...fileIndexMapping });
+
+    const writer = this.getAssetWriter({ index: SYNC_INDEX, isEager: true });
+    const fileWriter = this.getFileWriter({ index: FILE_SYNC_INDEX, isEager: true });
+    let offset = 0;
+    for (;;) {
```
wendlans:
```suggestion
    while (true) {
```
**EDIT:** here and anywhere else

apps/server-asset-sg/src/features/assets/search/asset-search-writer.service.ts
wendlans, line 1: What's up with the file naming? Shouldn't it be `asset-search-writer.service.ts`?
Same question for the other `writer.ts` file.

apps/server-asset-sg/src/features/assets/search/search.service.ts
```diff
@@ -0,0 +1,394 @@
...
+  /** Additional cardinality aggregations to compute alongside the search. */
+  countAggs?: Record<string, { field: string }>;
+}
+
+export interface SearchHitData {
+  fields?: Record<string, unknown[]>;
+  source?: Record<string, unknown>;
+  innerHits?: Record<
+    string,
+    Array<{
```
wendlans: Pure cosmetic: there is a wild mix between using `Array<type>` and `type[]`. Both are fine but I'd prefer one or the other throughout the project and I have a slight preference of the later.

apps/server-asset-sg/src/features/assets/search/search.service.ts
```diff
@@ -0,0 +1,394 @@
...
+    const { must, filter, aggs } = mapQueryToElasticDslParts(query, user);
+
+    const totalHitsResponse = await this.elastic.search({
+      index: this.index,
+      size: 0,
+      query: { bool: { must, filter } },
+      track_total_hits: true,
+      aggs,
+    });
+    const rawHitCount = (totalHitsResponse.hits.total as SearchTotalHits).value;
```
wendlans, line 131: Type assertions are always dangerous to use and should only be used if you're absolutely certain that the `type` is that and only that at that moment.
If you're not sure the cleaner way would be to check what kind of type `total` is here before accessing it.

apps/server-asset-sg/src/features/assets/search/search.service.ts
```diff
@@ -0,0 +1,394 @@
...
+      (totalHitsResponse.aggregations?.['distinct_assets'] as { value: number } | undefined)?.value ?? rawHitCount;
+    const totalFiles =
+      (totalHitsResponse.aggregations?.['distinct_files'] as { value: number } | undefined)?.value ?? 0;
+    const isNotAdminEdgecase = options?.unrestrictedWorkgroupQuery === undefined;
+    if (totalHits === 0 && isNotAdminEdgecase) {
+      return makeEmptyAssetSearchStats();
+    }
+
+    const makeAggregation = createMakeAggregation(query, user);
+    const aggregations = {
```
wendlans, line 144: There are a lot of magic strings/types in here.

apps/server-asset-sg/src/features/assets/search/search.service.ts
```diff
@@ -0,0 +1,394 @@
...
+    const aggResults = (
+      await this.elastic.search({
+        index: this.index,
+        size: 0,
+        query: { bool: { must } },
+        track_total_hits: true, // TODO: necessary?
+        aggregations,
+        filter_path: ['aggregations.*.a.buckets.*', 'aggregations.*.a.value'],
+      })
+    ).aggregations as unknown as NestedAggResult<AssetAggResult>;
```
wendlans: It's not the cleanest solution but better than a double assertion.
```suggestion
    ).aggregations as NestedAggResult<AssetAggResult> | undefined;
```

apps/server-asset-sg/src/features/assets/search/search.service.ts
```diff
@@ -0,0 +1,394 @@
...
+    elasticQuery: QueryDslQueryContainer,
+    page: PageOptions,
+    state: SearchState<EntityId>,
+    config?: SearchConfig,
+  ): Promise<boolean> {
+    const offset = page.offset ?? 0;
+    let remainingOffset = offset;
+    while (remainingOffset > 0) {
+      const remainingLimit = Math.min(SEARCH_BATCH_SIZE, remainingOffset);
+      if (remainingLimit <= 0) {
```
wendlans, line 202: This shouldn't be possible, right?

apps/server-asset-sg/src/features/assets/search/search.service.ts
```diff
@@ -0,0 +1,394 @@
...
+        config,
+      });
+      if (state.lastEntityId == null && state.totalCount != null && state.totalCount < offset) {
+        return false;
+      }
+      if (response.hits.hits.length < remainingLimit) {
+        return false;
+      }
+      remainingOffset -= response.hits.hits.length;
+      state.lastEntityId = response.hits.hits[response.hits.hits.length - 1].sort?.[0] as EntityId;
```
wendlans, line 217: Is the `as EntityId` necessary?

apps/server-asset-sg/src/features/assets/search/search.service.ts
```diff
@@ -0,0 +1,394 @@
...
+    return state.totalCount == null || state.totalCount > offset;
+  }
+
+  private async doSearchByOffset(
+    elasticQuery: QueryDslQueryContainer,
+    page: PageOptions,
+    state: SearchState<EntityId>,
+    config?: SearchConfig,
+  ) {
+    for (;;) {
```
wendlans:
```suggestion
    while(true) {
```

apps/server-asset-sg/src/features/assets/search/search.service.ts
```diff
@@ -0,0 +1,394 @@
...
+   * @param query The Elasticsearch query.
+   * @param page Pagination options (offset, limit).
+   * @param config Optional search configuration for controlling fields, sorting, and collapse.
+   */
+  public async search(
+    query: QueryDslQueryContainer,
+    page: PageOptions = {},
+    config?: SearchConfig,
+  ): Promise<PaginatedSearchResult<EntityId>> {
+    const state: SearchState<EntityId> = {
```
wendlans, line 94: Generally: I don't like the approach to use an object as argument of one or more methods in which it gets manipulated as a side effect. It's a hidden mechanic that requires you to have full knowledge on what's going on in detail.

apps/server-asset-sg/src/features/assets/search/search.service.ts
```diff
@@ -0,0 +1,394 @@
...
+      }
+      const response = await this.executeSearchQuery(elasticQuery, state, {
+        limit: remainingLimit,
+        fetchData: true,
+        config,
+      });
+      if (response.hits.hits.length === 0) {
+        return;
+      }
+      for (const hit of response.hits.hits) {
```
wendlans, line 242: There are a couple of type assertions within this for loop: are they safe?

apps/server-asset-sg/src/features/assets/search/search.service.ts
```diff
@@ -0,0 +1,394 @@
...
+   * Supports configurable sort field, field collapse, source/stored fields, and inner hits.
+   */
+  private async executeSearchQuery(
+    elasticQuery: QueryDslQueryContainer,
+    state: SearchState<EntityId>,
+    options: { limit: number; fetchData: boolean; config?: SearchConfig },
+  ): Promise<SearchResponse> {
+    const entityIdField = options.config?.entityIdField ?? 'id';
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    const searchParams: Record<string, any> = {
```
wendlans: Remove `// eslint-disable-next-line @typescript-eslint/no-explicit-any` and use the correct type here:
```suggestion
    const searchParams: SearchRequest = {
```

apps/server-asset-sg/src/features/assets/search/search.service.ts
```diff
@@ -0,0 +1,394 @@
...
+      track_total_hits: state.totalCount == null,
+    };
+
+    if (state.lastEntityId != null) {
+      searchParams['search_after'] = [state.lastEntityId];
+    }
+
+    // Collapse configuration for grouping results (e.g., pages by fileId).
+    if (options.config?.collapse) {
+      const collapseParam: Record<string, unknown> = {
```
wendlans:
```suggestion
      const collapseParam: SearchFieldCollapse = {
```

apps/server-asset-sg/src/features/assets/search/search.service.ts
wendlans: **EDIT**: you can replace all `object['xyz']` access with direct access `object.xyz` now

apps/server-asset-sg/src/features/assets/search/search.service.ts
```diff
@@ -0,0 +1,394 @@
...
+          ...(ih.highlight ? { highlight: { fields: ih.highlight, highlight_query: elasticQuery } } : {}),
+          ...(ih.sort ? { sort: ih.sort } : {}),
+        };
+      }
+      searchParams['collapse'] = collapseParam;
+    }
+
+    // Add cardinality aggregations on the first query for accurate counts.
+    if (state.totalCount == null) {
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
wendlans: Search for the correct type instead of using `any`.

apps/server-asset-sg/src/features/assets/sync/asset-sync.controller.ts
```diff
@@ -33,6 +33,21 @@ export class AssetSyncController {
...
+  async showFullTextState(@Res() res: Response): Promise<{ progress: number } | void> {
+    try {
+      const state = await this.fileFulltextSyncService.show();
+      if (state === null) {
+        res.status(204).end();
+        return;
+      }
+      res.status(200).json({ progress: state.progress }).end();
+    } catch (e) {
+      throw new HttpException(`${e}`, 500);
```
wendlans, line 47: General question: isn't it potentially dangerous to return the full error to the client? It could potentially contain some sensitive information, right?

apps/server-asset-sg/src/features/assets/sync/file-fulltext-sync.service.ts
```diff
@@ -52,14 +54,56 @@ export class FileFulltextSyncService extends AtomicProgressService<FileFulltextS
...
+
+    const total = await this.assetRepo.count();
+    if (total === 0) {
+      this.logger.debug('No assets to index files for');
+      await writeProgress(1);
+      return;
+    }
+
+    const fileWriter = this.searchWriterService.getFileWriter({
+      index: 'swissgeol_asset_file',
```
wendlans:
```suggestion
      index: FILE_ELASTIC_INDEX,
```

libs/asset-viewer/src/lib/components/asset-search-results/asset-search-results.component.html
```diff
@@ -87,6 +102,52 @@
           ></tr>
         </table>
       </div>
-    </div>
-  }
-</ng-container>
+    }
+    @if (searchQuery.type === SearchType.File) {
+      <div class="table-container">
```
wendlans, line 107: It would probably make sense to split this whole `File` part into its own separate component.

libs/asset-viewer/src/lib/components/asset-search-results/asset-search-results.component.ts
```diff
@@ -36,19 +49,29 @@ export class AssetSearchResultsComponent implements OnInit, OnDestroy {
     'createDate',
   ];
 
+  protected readonly FILE_COLUMNS = ['fileName', 'assetTitle', 'pages', 'actions'];
+
+  public activeTab: 'assets' | 'files' = 'assets';
```
wendlans: Can be removed. This is only used in `setActiveTab` which is unused.

libs/asset-viewer/src/lib/components/asset-search-results/asset-search-results.component.ts
```diff
@@ -79,6 +102,10 @@ export class AssetSearchResultsComponent implements OnInit, OnDestroy {
     }
   }
 
+  public setActiveTab(tab: 'assets' | 'files'): void {
```
wendlans: This is currently unused.

libs/asset-viewer/src/lib/components/asset-viewer-page/asset-viewer-page.component.ts
```diff
@@ -89,12 +74,7 @@ export class AssetViewerPageComponent implements OnInit, OnDestroy {
   public hasCurrentAsset$ = this.store.select(fromAppShared.selectHasCurrentAsset);
 
   public isFiltersOpen$ = this.store.select(selectIsFiltersOpen);
-
-  public searchTextKeyDown$ = new Subject<KeyboardEvent>();
-  private readonly searchTextChanged$ = this.searchTextKeyDown$.pipe(
-    filter((event) => event.key === 'Enter'),
-    map((event) => (event.target as HTMLInputElement).value),
-  );
+  public searchQuery$ = this.store.select(selectSearchQuery).pipe(map((q) => q.text ?? ''));
```
wendlans: unused

libs/asset-editor/src/lib/components/asset-editor-sync/asset-editor-sync.component.ts
```diff
@@ -27,21 +33,43 @@ export class AssetEditorSyncComponent implements OnInit {
...
+    await this.loopSyncProgress();
   }
 
-  private async loopAssetSyncProgress() {
+  async synchronizeFilesToElastic() {
+    if (this.fileSyncProgress !== null) {
+      return;
+    }
+    this.fileSyncProgress = 0;
+    await resolveFirst(this.httpClient.post('/api/assets/sync/full-text?reloadFromS3=true', null));
```
wendlans: Why is this component accessing the web server directly while others are using a dedicated service to do so? (e.g. `asset-search.service.ts`)

libs/asset-viewer/src/lib/services/viewer-controller.service.ts
```diff
@@ -135,28 +143,43 @@ export class ViewerControllerService {
...
   private async loadResults(
-    query: AssetSearchQuery,
+    query: SearchQueries,
     options: { force?: boolean; skipAssetReset?: boolean } = {},
   ): Promise<void> {
     // Always load results when favoritesOnly is true, even if other search criteria are empty
     if (!options.force && isEmptySearchQuery(query) && !query.favoritesOnly) {
-      this.store.dispatch(actions.setResults({ results: makeEmptyAssetSearchResults(), isLoading: false }));
+      this.store.dispatch(actions.setAssetsResults({ results: makeEmptyAssetSearchResults(), isLoading: false }));
+      this.store.dispatch(actions.setFileResults({ fileResults: makeEmptyFileSearchResults(), isLoading: false }));
```
wendlans: To use multiple dispatches is an anti-pattern with NgRx and can usually be solved by creating a new action that combines both.

libs/asset-viewer/src/lib/services/viewer-controller.service.ts
```diff
@@ -135,28 +143,43 @@ export class ViewerControllerService {
...
     }
-    this.store.dispatch(actions.setResults({ isLoading: true }));
-    const results = await firstValueFrom(this.assetSearchService.search(query));
-    this.store.dispatch(actions.setResults({ results, isLoading: false }));
-    if (results.data.length === 1) {
-      await this.loadAsset(results.data[0].id);
-    } else if (!options.skipAssetReset) {
-      this.store.dispatch(appSharedStateActions.setCurrentAsset({ asset: null, isLoading: false }));
+
+    if (query.type === SearchType.File) {
```
wendlans: Use a `switch-case` here; it's better if additional search types will be added in the future.

libs/asset-viewer/src/lib/services/viewer-controller.service.ts
```diff
@@ -135,28 +143,43 @@ export class ViewerControllerService {
...
     }
   }
 
-  private async loadStats(query: AssetSearchQuery): Promise<void> {
+  private async loadStats(query: SearchQueries): Promise<void> {
     this.store.dispatch(actions.setStats({ isLoading: true }));
-    const stats = await firstValueFrom(this.assetSearchService.searchStats(query));
-    this.store.dispatch(actions.setStats({ stats, isLoading: false }));
+    let stats: AssetSearchStats | undefined;
+    if (query.type === SearchType.File) {
```
wendlans: Same here: use a `switch-case` instead

libs/client-shared/src/lib/components/search-input/search-input.component.ts
```diff
@@ -0,0 +1,24 @@
...
+import { SvgIconComponent } from '@ngneat/svg-icon';
+import { TranslatePipe } from '@ngx-translate/core';
+
+@Component({
+  selector: 'asset-sg-search-input',
+  imports: [SvgIconComponent, TranslatePipe, FormsModule],
+  templateUrl: './search-input.component.html',
+  styleUrl: './search-input.component.scss',
+})
+export class SearchInputComponent {
```
wendlans, line 12: +1 for using Signals ❤️

libs/client-shared/src/lib/components/search-input/search-input.component.ts
```diff
@@ -0,0 +1,24 @@
...
+  readonly value = model('');
+  readonly placeholderTranslationKey = input('search.textSearchFieldPlaceholder');
+  readonly keydownEvent = output<KeyboardEvent>();
```
wendlans, line 15: unused

development/init/elasticsearch/mappings/swissgeol_asset_file.json
```diff
@@ -0,0 +1,102 @@
+{
+  "properties": {
+    "id": {
+      "type": "keyword"
+    },
+    "assetId": {
+      "type": "keyword"
```
wendlans, line 7: Shouldn't that be an `integer`? Same question with `fileId`.

libs/client-shared/src/lib/components/toggle-button/toggle-button.component.html
```diff
@@ -0,0 +1,12 @@
+<button asset-sg-tertiary [class.active]="!isActive" (click)="onToggleIsActive()" translate>
+  @if (inactiveTranslationIcon) {
+    <svg-icon [key]="inactiveTranslationIcon"></svg-icon>
+  }
+  {{ inactiveTranslationKey }}
```
wendlans, line 5: Why isn't this translated but the active one is?

apps/server-asset-sg/src/features/assets/files/file-processors/file-ocr/file-ocr.service.ts
```diff
@@ -49,6 +53,9 @@ export class FileOcrService extends AbstractProcessingService<[]> {
       where: { id: file.id },
     });
 
+    await this.fileService.loadFulltextContentFromS3(file.id);
```
wendlans: This runs before the `fileProcessingState` is checked for errors. Is that correct in case of errors?

apps/server-asset-sg/src/features/assets/search/search-writer.service.ts
```diff
@@ -0,0 +1,156 @@
...
+  getFileWriter(options?: SearchWriterOptions): FileSearchWriter {
+    return new FileSearchWriter(
+      this.elastic,
+      this.prisma,
+      this.geometryRepo,
+      options ?? { index: FILE_ELASTIC_INDEX, shouldRefresh: true },
+    );
+  }
+
+  async deleteFromIndex(assetId: AssetId): Promise<void> {
```
wendlans, line 56: My friend Claude found some issues here. Not sure if that's actually an issue. Also not sure if the solution is best practise.
> ### `deleteFromIndex` will crash on missing ES documents
> 
> ```typescript
> // search-writer.service.ts
> async deleteFromIndex(assetId: AssetId): Promise<void> {
>   await Promise.all([
>     this.elastic.delete({           // ← throws 404 if doc doesn't exist!
>       index: ASSET_ELASTIC_INDEX,
>       id: `${assetId}`,
>       refresh: true,
>     }),
>     this.elastic.deleteByQuery({    // ← graceful (ignore_unavailable)
>       index: FILE_ELASTIC_INDEX,
>       ...
>     }),
>   ]);
> }
> ```
> 
> Called from `AssetService.delete`:
> ```typescript
> const [ok] = await Promise.all([
>   this.assetRepo.delete(id),          // DB delete
>   this.searchWriterService.deleteFromIndex(id)  // ES delete
> ]);
> ```
> 
> **Problem**: If the asset doesn't exist in ES (e.g., first deployment, failed sync, test
> environment), `elastic.delete()` throws a 404 `ResponseError`. Since both operations are in
> `Promise.all`, the database delete may have already started but the caller receives an error.
> The asset would be deleted from the DB but the error propagates, possibly triggering a 500
> response to the client.
> 
> **Fix**:
> ```typescript
> this.elastic.delete({ ... }).catch((e) => {
>   if (e?.meta?.statusCode !== 404) throw e;
> })
> ```

apps/server-asset-sg/src/features/assets/search/file-search.writer.ts
```diff
@@ -0,0 +1,369 @@
...
+      fetchSharedEagerData(this.prisma, this.geometryRepo),
+      this.fetchEagerFulltextContent(),
+    ]);
+    return {
+      ...sharedData,
+      fileIdToFulltextContent: fulltextContent,
+    };
+  }
+
+  private async fetchEagerFulltextContent(): Promise<Map<number, { fileName: string; pages: FulltextContent[] }>> {
```
wendlans: Claude found some issues here:

> ### Eager Fulltext Loading Can OOM the Process
> 
> ```typescript
> // file-search.writer.ts → fetchEagerFulltextContent()
> const files = await this.prisma.file.findMany({
>   where: { fulltextContent: { not: Prisma.AnyNull } },
>   select: { id: true, name: true, nameAlias: true, fulltextContent: true },
> });
> ```
> 
> This loads **ALL fulltext content into memory at once** during `syncWithDatabase` and
> `FileFulltextSyncService.sync` (both use `isEager: true`).
> 
> Rough calculation: 10,000 files × 50 pages × 3KB/page = **~1.5 GB** in a single Node.js Map.
> With larger datasets this will exceed the default V8 heap limit and crash.
> 
> The eager pattern works fine for contacts (~100 records) and favorites (small), but fulltext
> content is orders of magnitude larger.
> 
> **Recommendation**: Stream fulltext content in batches rather than loading all at once. Or skip
> eager loading for fulltext (the `fetchFulltextContentForAsset` per-asset path already exists and
> only loads what's needed).

apps/server-asset-sg/src/features/assets/search/search-query.utils.ts
```diff
@@ -0,0 +1,250 @@
...
+              },
+            ],
+          },
+        };
+        break;
+      case SearchType.File:
+        textQuery = {
+          match: {
+            content: {
+              query: escapeElasticQuery(normalizeFieldQuery(query.text)),
```
wendlans: Another thing that Claude found:

> ### `escapeElasticQuery` + `normalizeFieldQuery` Applied to File `match` Queries
> 
> `escapeElasticQuery` escapes ES query_string syntax characters (`(`, `)`, `!`, `+`, etc.) with
> backslashes. This makes sense for `query_string` (asset search) but is **wrong for `match`
> queries** (file search). The `match` query uses the field's analyzer, not query_string syntax.
> 
> Effects:
> - User searches "C++ programming" → escaped to `"C\+\+ programming"` → analyzer tokenizes
>   backslashes away → probably still works, but behavior is analyzer-dependent
> - `normalizeFieldQuery` converts `title:something` to field query syntax → meaningless in a
>   `match` on `content` → searches for pages literally containing "title" and "something"
> 
> **Recommendation**: Only apply `escapeElasticQuery` for `SearchType.Asset`. For file search,
> pass the raw text (or apply only basic sanitization).

apps/server-asset-sg/src/features/assets/search/search-query.utils.ts
```diff
@@ -0,0 +1,250 @@
...
+ * that contains any of a given set of values (the _query_).
+ *
+ * When the query is empty, only documents for which the field does not exist are matched.
+ * A field does also count as "not existing" when it is set to `null` or an empty array.
+ *
+ * @param field The field to match.
+ * @param query The set of allowed values.
+ */
+export const makeArrayFilter = <T extends string | number>(
+  field: keyof ElasticsearchAsset,
```
wendlans, line 162: Not really a problem but this is currently used for both `ElasticsearchAsset` and `ElasticsearchFilePage` searches.

apps/server-asset-sg/src/features/assets/search/search-writer.service.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  async count(): Promise<number> {
+    return (await this.elastic.count({ index: ASSET_ELASTIC_INDEX, ignore_unavailable: true })).count;
+  }
+
+  async countFiles(): Promise<number> {
+    return (await this.elastic.count({ index: FILE_ELASTIC_INDEX, ignore_unavailable: true })).count;
+  }
+
+  async syncWithDatabase(onProgress?: (percentage: number) => void | Promise<void>): Promise<void> {
```
wendlans, line 80: What happens if the process runs in some errors in the middle of syncing? The temporary sync indices will stay forever then, right?

apps/server-asset-sg/src/features/assets/sync/file-fulltext-sync.service.ts
```diff
@@ -52,14 +54,56 @@ export class FileFulltextSyncService extends AtomicProgressService<FileFulltextS
...
+    const progressScale = options?.reloadFromS3 ? 0.5 : 1;
+
+    const total = await this.assetRepo.count();
+    if (total === 0) {
+      this.logger.debug('No assets to index files for');
+      await writeProgress(1);
+      return;
+    }
+
+    const fileWriter = this.searchWriterService.getFileWriter({
```
wendlans, line 69: Another find by Monsieur Claude:

> ### `FileFulltextSyncService.sync()` Doesn't Clear Existing File Index
> 
> Unlike `syncWithDatabase()` (which recreates indices from scratch), the file-only sync in
> `FileFulltextSyncService.sync()` writes directly to the live file index without clearing it first:
> 
> ```typescript
> const fileWriter = this.searchWriterService.getFileWriter({
>   index: 'swissgeol_asset_file',
>   isEager: true,
> });
> // ... iterates assets, calls fileWriter.writeAssetFiles(records)
> ```
> 
> `writeAssetFiles` does `deleteByQuery` per batch of assets before re-indexing, but if an asset
> was deleted from the DB between syncs, its orphaned file pages remain in the index forever.
> 
> **Recommendation**: Either delete the entire file index before sync, or add a cleanup step
> afterward that removes docs for assets no longer in the DB.

wendlans (changes requested):
Woah... that was a **lot** of code... 😖
Anyway, I've added comments to all sections that you should review. I think most SonarCloud comments are correct and should be fixed as well.

.github/workflows/publish-prod.yml
vej-ananas: This is intentional - the change was already applied to master and for some reason went into this branch rather than directly into develop.

apps/server-asset-sg/src/features/assets/search/file-search.service.ts
vej-ananas: I think not.

apps/server-asset-sg/src/features/assets/search/file-search-writer.service.ts
vej-ananas: This is consistent with the approach in existing services - and it is pretty neat, no?

apps/server-asset-sg/src/features/assets/search/search.service.ts
vej-ananas: I agree. but i did neither write nor touch the offset method - it was written before.

apps/server-asset-sg/src/features/assets/search/search.service.ts
vej-ananas: This is pre existing code.

apps/server-asset-sg/src/features/assets/search/search.service.ts
vej-ananas: This is pre existing code.

apps/server-asset-sg/src/features/assets/sync/asset-sync.controller.ts
vej-ananas: It seems to be current practice.

development/init/elasticsearch/mappings/swissgeol_asset_file.json
vej-ananas: that probabtly depends on how we want to search for them, right?

libs/asset-editor/src/lib/components/asset-editor-sync/asset-editor-sync.component.ts
vej-ananas: I do not know.

libs/asset-viewer/src/lib/components/asset-search-results/asset-search-results.component.html
vej-ananas: So two new components for each table? Or how would you separate?

Imo the file size is still okay.

libs/client-shared/src/lib/components/toggle-button/toggle-button.component.html
vej-ananas: in fact they are both translated, see the translate attribute in the button :) but the `| translate` is obsolete.

apps/server-asset-sg/src/features/assets/search/search-query.utils.ts
vej-ananas: Not that I know of.

apps/server-asset-sg/src/features/assets/search/file-search-writer.service.ts
vej-ananas: It is not defined afaik.

apps/server-asset-sg/src/features/assets/search/search-writer.service.ts
vej-ananas: I dont see this as an issue. If we experience problems, we can look into this.

apps/server-asset-sg/src/features/assets/sync/asset-sync.controller.ts
vej-ananas: We could take this into a maintenance issue.

apps/server-asset-sg/src/features/assets/search/file-search-writer.service.ts
wendlans: Schade eigentlich. I'll throw it on the technical dept pile.
#926

apps/server-asset-sg/src/features/assets/search/file-search-writer.service.ts
wendlans: Consistency should always have higher prio 👌
And yes, it's kinda neat but I usually prefer clean interfaces/types. However, I can see the appeal of this.

apps/server-asset-sg/src/features/assets/search/search-query.utils.ts
wendlans: Another thing for the technical dept pile 😅

apps/server-asset-sg/src/features/assets/sync/asset-sync.controller.ts
wendlans: I've quickly seached for other places where that happens and found a very few. Nonetheless, I've added a new issue: #927

libs/asset-viewer/src/lib/components/asset-search-results/asset-search-results.component.html
wendlans: It's not just about the file size; it's also a style decision so that each component has a single concept.
And yes, each type (assets, files) would get their own component.

libs/asset-editor/src/lib/components/asset-editor-sync/asset-editor-sync.component.ts
wendlans: Is there a way to make this more consistent? Like encapsulating this in a dedicated service? It would help if we want to change something on the API.

development/init/elasticsearch/mappings/swissgeol_asset_file.json
wendlans: Right now it's a mix between number and string in the code. That's why I'm asking. However, I don't know ElasticSearch.

wendlans (commented):
Thanks a lot for all the updates. There are only a few comments left. We can have a look at this tomorrow.

apps/server-asset-sg/src/features/assets/sync/asset-sync.controller.ts
wendlans: As discussed: You can fix it directly within this PR or later as part of the linked issue.

development/init/elasticsearch/mappings/swissgeol_asset_file.json
vej-ananas: It should be fine.

wendlans (approved):
Thanks for the effort to work through all the comments and fixing so many of them 👍

PR #937, opened 2026-05-06
#794 Seitenklassifikation: Metadaten anzeigen

Dead0utside:
- Updates the file-extraction-service API to V2;
- Updates Data Models to store the added optional title for page range classifications
- Update UI to display the title if it exists
- Deprecate file-extraction-service API V1


libs/client-shared/src/lib/components/pdf-viewer/pdf-viewer-toc/pdf-viewer-toc.component.html
```diff
@@ -20,7 +20,7 @@
                 <span
                   class="pdf-viewer-toc__content__item__content__category"
                   [class.pdf-viewer-toc__content__item__content__category--active]="isActive"
-                  >{{ categoryTitle }} {{ $index + 1 }}</span
+                  >{{ classification.label ?? categoryTitle + "" + ($index + 1) }}</span
```
uhon: Wurde das vorher nicht als "Leerzeichen" gerendert?

```suggestion
                  >{{ classification.label ?? categoryTitle + " " + ($index + 1) }}</span
```

apps/server-asset-sg/src/features/assets/files/file-processors/file-extraction/file-extraction.service.ts
```diff
@@ -59,9 +61,9 @@ export class FileExtractionService extends AbstractProcessingService<PredictionS
     super();
   }
 
-  @OnEvent(EVENTS.FILE_START_EXTRACT)
-  protected async handleStartEvent(payload: ProcessableFile): Promise<void> {
-    void this.process(payload);
+  /** @deprecated Handled by v2/FileExtractionService. Remove once v2 is confirmed stable. */
```
uhon, line 64: Wouldn't it make sense to keep the old code (`this.process(payload`) and just disable the `@OnEvent` trigger by commenting it out.

apps/server-asset-sg/src/features/assets/asset.module.ts
```diff
@@ -66,6 +67,7 @@ import { UsersModule } from '@/features/users/users.module';
     FileS3Service,
     FileService,
     FileExtractionService,
+    FileExtractionServiceV2,
```
uhon: Why is the old FileExtractionService still provided? Does other code use the old version?

apps/server-asset-sg/src/features/assets/files/file-processors/file-extraction/v2/file-extraction.service.ts
```diff
@@ -0,0 +1,144 @@
...
+      entityCount: result?.entities?.length ?? 0,
+    });
+
+    if (result == null || !Array.isArray(result.entities)) {
+      this.logger.error('[EXTRACT-V2] Extraction API returned no valid result.');
+      await this.updateStatus(file, FileProcessingState.Error);
+      return;
+    }
+
+    const ranges = this.mapEntitiesToPageRanges(result.entities);
```
uhon, line 75: what if result.entities is empty? maybe a log message would make sense.

uhon (changes requested):
I found some small issues.
Maybe it deserves its onw test, since v1 was covered by tests.

apps/server-asset-sg/src/features/assets/asset.module.ts
Dead0utside: I don't think so but I decided to keep it as an easy fallback option in case V2 needs further adjustments

apps/server-asset-sg/src/features/assets/files/file-processors/file-extraction/file-extraction.service.ts
Dead0utside: commenting code out is generally discouraged and this functionality is deprecated anyway. I believe it's fine to keep it the way it is, and remove the deprecated service altogether as soon as the new functionality is tested and works

uhon: I thought if we would have the need to revert to V1, the service wouldn't process the payload anymore, because of the removal of `this.process(payload)`, but we also don't want V1 and V2 to react to the event in parallel.
So why not leave the code intact and just remove it from asset.module.ts and re-introduce it again in case we need to revert to V1?

Dead0utside: good idea, I implemented it

uhon (approved):
LGTM now

PR #967, opened 2026-05-29
#794: Seitenklassifikation ergaenzungen

Dead0utside:


apps/client-asset-sg/src/app/i18n/fr.ts
```diff
@@ -221,13 +221,16 @@ export const frTranslationMapping: AppTranslationMapping = {
...
+          updateOrder: "Actualiser l'ordre",
           type: 'Type',
           language: 'Langue',
+          label: 'Titre',
+          addSection: 'Ajouter une section',
           pageFrom: 'Page de',
           pageTo: 'Page à',
           save: 'Terminer la modification',
           toLessThanFromError: "La 'Page de' doit être inférieure ou égale à la 'Page à'.",
+          labelTooLong: 'Le titre ne doit pas dépasser 120 caractères.',
```
uhon: This translation is never used in code. Are we missing a validator impl?

libs/asset-editor/src/lib/components/asset-editor-tabs/asset-editor-files/page-range-editor/page-range-editor.component.ts
```diff
@@ -191,8 +242,21 @@ export class PageRangeEditorComponent {
           Validators.required,
           Validators.minLength(1),
         ]),
+        label: this.fb.control(pc.label ?? null),
```
uhon, line 245: do we need to apply the maxLength validator here?

libs/asset-viewer/src/lib/components/asset-viewer-files/asset-viewer-files-content/asset-viewer-files-content.component.html
```diff
@@ -29,7 +29,7 @@
           </dt>
           @for (classification of group.classifications; track classification) {
             <dd>
-              <button (click)="this.openPdfOnPage(classification.from)" [title]="classification.label">
+              <button (click)="this.openPdfOnPage(classification.from)" [attr.title]="classification.label ?? null">
```
uhon: What if the classification.label is `""` (can this even be the case?)
It would add a attribute `title=""`

package-lock.json
uhon, line 1: no changes happend to package.json. I think changes to this file are irrelevant?

uhon (changes requested):
Minor issues found. Newly introduced Translation Key is not used.

package-lock.json
Dead0utside: the changes to the file happened after executing `npm install`. I would assume that some of the internal dependencies may have changed but it's probably safe to discard the updates of this file

libs/asset-viewer/src/lib/components/asset-viewer-files/asset-viewer-files-content/asset-viewer-files-content.component.html
Dead0utside: I don't think it's possible but good point, this should be handled

libs/asset-editor/src/lib/components/asset-editor-tabs/asset-editor-files/page-range-editor/page-range-editor.component.ts
Dead0utside: I had used the validator initially but couldn't find the requirement, so ultimately decided to get rid of it. Do you think we need to apply the maxLength validator nevertheless?

apps/client-asset-sg/src/app/i18n/fr.ts
Dead0utside: this is also a translation label for the validator I've ultimately decided to not use. Can be removed unless we decide to use the length limit validation for the title

uhon (commented):
LGTM now. Thanks!

uhon (approved):
LGTM, now with "Approve" flag

PR #969, opened 2026-06-03
#965 layout fixes for search results.

uhon:
Limits width of long texts and clamp height to max 3 lines.
Also moves info-button to the front of document search.
After discussion with @mPfifi , we don't want to strictly avoid horizontal scroll.

### DIFF
--- a/libs/asset-viewer/src/lib/components/asset-search-results/asset-search-results.component.html
+++ b/libs/asset-viewer/src/lib/components/asset-search-results/asset-search-results.component.html
@@ -35,7 +35,11 @@
           </ng-container>
           <ng-container matColumnDef="titlePublic">
             <th mat-header-cell *matHeaderCellDef translate>{{ "edit.tabs.references.assetTitlePublic" }}</th>
-            <td mat-cell *matCellDef="let asset">{{ asset.title }}</td>
+            <td mat-cell *matCellDef="let asset">
+              <div class="clamped-cell-content">
+                {{ asset.title }}
+              </div>
+            </td>
           </ng-container>
           <ng-container matColumnDef="assetFormat">
             <th mat-header-cell *matHeaderCellDef translate>{{ "search.kind" }}</th>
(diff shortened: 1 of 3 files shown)


libs/asset-viewer/src/lib/components/asset-search-results/asset-search-results.component.html
```diff
@@ -106,13 +116,31 @@
     @if (searchQuery.type === SearchType.File) {
       <div class="table-container">
         <table mat-table [dataSource]="fileResultsToDisplay" class="mat-elevation-z8 table">
+          <ng-container matColumnDef="actions">
```
vej-ananas, line 119: Just fyi: making this change orally and not via Simon and Figma will result in Figma and the App being out of Sync. I am not entirely sure what the correct procedure is - maybe something to clarify with Simon.

libs/asset-viewer/src/lib/components/asset-search-results/asset-search-results.component.scss
```diff
@@ -73,8 +73,17 @@
       height: 40px;
     }
 
-    .mat-column-titlePublic {
+    .clamped-cell-content {
+      line-clamp: 3;
+      overflow: hidden;
+      display: -webkit-box;
+      -webkit-box-orient: vertical;
+      -webkit-line-clamp: 3;
```
vej-ananas, line 80: According to https://modern-css.com/multiline-text-truncation-without-javascript/, you can instead use `line-clamp: 3;` and remove `-webkit-box-orient: vertical;`

vej-ananas (approved):

libs/asset-viewer/src/lib/components/asset-search-results/asset-search-results.component.scss
uhon: I removed `-webkit-box-orient: vertical;` seems unnecessary. left the other webkit properties for backwards compatibility

libs/asset-viewer/src/lib/components/asset-search-results/asset-search-results.component.html
uhon: I told Simon about the change location of the info-icon. He can decide whether to update it in Figma.

PR #980, opened 2026-06-09
Bug/assets 977 document indexing failures after pre release

uhon:
Avoids single files crashing the indexing process.
Improves logging so we know what is going on.
Removes clutter (Font-Loading warnings) from Logs
Avoids crashing the NodeJS process if elastic imports fail (like exceeded Memory)

### DIFF
--- a/apps/server-asset-sg/src/features/assets/files/file.service.ts
+++ b/apps/server-asset-sg/src/features/assets/files/file.service.ts
@@ -222,7 +222,7 @@ export class FileService {
         data: { fulltextContent: content as unknown as Prisma.JsonArray },
       });
     } catch (e) {
-      this.logger.warn('Failed to load fulltext content from S3', { fileId, error: e });
+      // this.logger.warn('Failed to load fulltext content from S3', { fileId, error: e });
     }
   }
 
@@ -247,7 +247,7 @@ export class FileService {
         yield { pageNumber: i, text };
       }
     } finally {
-      await doc?.loadingTask.destroy();
+      doc?.destroy();
     }
   }
 
--- a/apps/server-asset-sg/src/features/assets/files/file.service.ts
+++ b/apps/server-asset-sg/src/features/assets/files/file.service.ts
@@ -222,7 +222,7 @@ export class FileService {
         data: { fulltextContent: content as unknown as Prisma.JsonArray },
       });
     } catch (e) {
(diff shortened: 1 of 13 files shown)


apps/server-asset-sg/src/features/assets/files/pdf-metadata.service.ts
```diff
@@ -75,7 +83,7 @@ export class PdfMetadataService {
       throw new Error(`PDF metadata extraction failed: ${error instanceof Error ? error.message : 'Unknown error'}`);
     } finally {
       // Ensure PDF document is properly destroyed to free memory
-      await doc?.destroy();
+      doc?.destroy();
```
vej-ananas: Why not await the promise?

apps/server-asset-sg/src/features/assets/files/file.service.ts
```diff
@@ -215,7 +222,7 @@ export class FileService {
         data: { fulltextContent: content as unknown as Prisma.JsonArray },
       });
     } catch (e) {
-      this.logger.warn('Failed to load fulltext content from S3', { fileId, error: e });
+      // this.logger.warn('Failed to load fulltext content from S3', { fileId, error: e });
```
vej-ananas: Dont comment out code.

apps/server-asset-sg/src/features/assets/files/file.service.ts
```diff
@@ -27,6 +28,12 @@ import { CreateFileData, FileIdentifier, FileRepo } from '@/features/assets/file
 import { PdfMetadataService } from '@/features/assets/files/pdf-metadata.service';
 import { mapAssetLanguagesToPrismaUpdate } from '@/features/assets/prisma-asset';
 
+// eval('require') bypasses webpack's static require analysis so the path is resolved at runtime by Node.js.
+
+const STANDARD_FONT_DATA_URL =
```
vej-ananas, line 33: Would something like `pdfjs-dist/standard_fonts` not be easier to understand? Not sure if that works though..

vej-ananas (approved):

apps/server-asset-sg/src/features/assets/files/file.service.ts
uhon: Will try, everything else fixed

vej-ananas (approved):

PR #990, opened 2026-06-10
Fix PDF-Viewer, which had issues because of unwanted version bump

uhon:
pdfjs-dist fixed at version "5.4.530" instead of using "^".
This fixes PDF-Viewer. re-applied standardFontDataUrl workaround. It was working as expected before


Dead0utside (approved):
lgtm

PR #992, opened 2026-06-10
Cancel long running file indexing processes

uhon:
Circumvent that corrupt or large PDF files can stall the pdfjs-dist extraction.

We race the pdfjs-dist promise against a timer which rejects the promise if it takes longer than 120s

To prevent version mismatches between bundled pdfjs and cached pdfjs-web-worker we reference the web-worker file by adding the correct version of it as query-parameter (v). -> Needs adjustiment of nginx config, which resolves files with a query-string incorrectly (concatenates path with query string), which worked so far, since none of the static files contained a queryString. Now we can use the query-string just as a cache-buster. Might become handy for future updates of pdf-js to circumvent stale web-workers

### DIFF
--- a/apps/client-asset-sg/docker/nginx.conf
+++ b/apps/client-asset-sg/docker/nginx.conf
@@ -54,6 +54,6 @@ server {
         }
 
         expires $expires;
-        try_files $uri$args $uri$args/ /index.html;
+        try_files $uri $uri/ /index.html;
     }
 }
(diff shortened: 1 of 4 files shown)


apps/client-asset-sg/docker/nginx.conf
```diff
@@ -54,6 +54,6 @@ server {
         }
 
         expires $expires;
-        try_files $uri$args $uri$args/ /index.html;
```
vej-ananas, line 57: is this intentional?

vej-ananas (approved):

PR #996, opened 2026-06-15
Feature/assets 790 fix file search issues

vej-ananas:


apps/server-asset-sg/src/features/assets/files/file.service.ts
```diff
@@ -187,10 +190,23 @@ export class FileService {
         await writeProgress?.(Math.min(i / files.length, 1));
       }
       i++;
-      await this.loadFulltextContentFromS3(file.id);
+      try {
```
uhon: I don't see how this should help. We already have a Catch-Block in `loadFulltextContentFromS3` which does not re-throw the error. So it should just continue with the loop anyways. Log below would never appear.

apps/server-asset-sg/src/features/assets/search/file-search-writer.service.ts
```diff
@@ -274,11 +282,19 @@ export class FileSearchWriterService {
...
         index: this.options.index,
         refresh: this.options.shouldRefresh,
         operations: chunk,
       });
+      if (result.errors) {
+        const failed = result.items.filter((item) => item.index?.error);
+        this.logger.error('Bulk indexing had errors', {
+          failedCount: failed.length,
+          totalCount: result.items.length,
+          errors: failed.slice(0, 5).map((item) => item.index?.error),
```
uhon, line 295: Why only show 5?

apps/server-asset-sg/src/features/assets/search/search-writer.service.ts
```diff
@@ -138,25 +146,58 @@ export class SearchWriterService {
...
+        this.logger.error('Failed to delete file sync index', { index: FILE_SYNC_INDEX, error: e });
+      });
+    }
+  }
+
+  /**
+   * Starts a reindex operation asynchronously and polls until it completes.
+   * This avoids HTTP timeouts on large datasets.
+   */
+  private async waitForReindex(sourceIndex: string, destIndex: string): Promise<void> {
```
uhon: I think naming is confusing here. It should be called differently. it does not only wait for reindex but is actually doing a reindex

uhon (changes requested):
Great updates on the elastic side as well as Unicode-Parsing behavior,

Please see Naming issue and Out-Of-Reach logs.

uhon (approved):

PR #1006, opened 2026-06-16
fix: isolate integration tests from development database

uhon:
Tests in server-asset-sg use TRUNCATE on all tables, which wipes the
 development database when triggered (e.g. by AI coding agents running
 tests to verify changes).

 - Add dedicated postgres_test database for test isolation
 - Add .env.test with test DATABASE_URL
 - Add jest globalSetup to auto-apply migrations to test DB
 - Load .env.test with override in jest.setup.ts
 - Add safety guard in test/setup-db.ts refusing to clear non-test DBs

### DIFF
--- a/.env.test
+++ b/.env.test
@@ -1,0 +1,1 @@
+DATABASE_URL=postgres://postgres:postgres@localhost:5432/postgres_test?schema=public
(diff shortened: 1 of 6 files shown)


vej-ananas (approved):
Are these tests being run somewhere in CI/CD and would this make problems?

PR #1008, opened 2026-06-18
Fix file indexing

vej-ananas:
#977 
pdfjs-dist leaks internal fetch rejections that bypass try/catch and crash Node.js.
We catch them via global error handling.


uhon (approved):
Wow, we seem not the only ones affected by this issue and it is still unsolved in PDFJS 6.

It look good.
To follow DRY-principle and avoid including it in future usages of `getDocument`, we could evaluate providing this mechanism as a separte function wrapping PDFJS tasks

PR #1017, opened 2026-06-25
Feature/assets 970 pdf viewer zoom and scroll improvements

uhon:
Implements CTR+Scroll functionality with seamless in place zoom adjustments and rerndering of pages in correct resolution.

Features/Problems-Solved:

-  OCR cache invalidation – After OCR processing (which may change page dimensions) the cached pageDimensions in the DB are now invalidated. Next PDF load re-derives fresh dimensions from the updated file.
- custom initial zoom level –  Independent of what zoom 1 (wides page fits) means, we apply an initial zoom-level to fit the tallest page into the vertical viewport, unless page gets even smaller because of very wide pages. Reset-Button resets this initial value
 - Text layer rotates with the document and re-calculates bounding-boxes for selection correctly
 - CTRL+wheel zoom via CSS transform – Rapid pinch/scroll zoom applies an instant CSS scale() on the pages container  (no TanStack recalculation per tick). After a 300ms settle delay, the zoom is committed: layout recalculated, scroll position anchored to cursor, re-render triggered.
 - Handover layer for seamless zoom transitions – On zoom commit (the 300ms from above), existing canvases are reparented into a temporary overlay positioned at the target layout. Pages crossfade out as fresh renders arrive, eliminating blank flashes.
 - Rapid zoom resilience – Re-entering the CSS-transform phase cancels in-flight renders from the previous commit. The render continues in the background while the transform provides visual feedback. Transition to 'normal' mode is also guarded while a new wheel gesture is pending.
 - Button zoom scroll anchoring – +/- zoom now preserves the vertical position within the anchor page and the horizontal scroll ratio, preventing disorienting jumps.


vej-ananas (commented):
## PDF Viewer Zoom & Scroll Improvements — Review

**Summary:** Substantial enhancement (988 additions) adding CTRL+wheel zoom with CSS-transform preview, a handover layer for smooth visual transitions, rotation scroll anchoring, text-layer rotation support, and initial zoom-to-fit. The architecture is well-decomposed across services.

---

### :red_circle: Issues

**1. Nested `setTimeout` chains in `releasePage` are not cleaned up on `end()`**

In `pdf-viewer-handover.service.ts:141-151`, `releasePage` creates nested `setTimeout` callbacks. If `end()` is called while these timers are in flight (e.g. from `begin()` → `this.end()`), the outer/inner timeouts will still fire and call `pageEl.remove()` on already-removed elements and check `this.releasedViewportPages` / `this.viewportPageNums` which were `.clear()`ed. This is a latent bug — harmless today (orphan DOM removal is safe) but the stale state checks could produce surprising behavior with rapid zoom gestures.

**2. Commented-out debug overlays left in SCSS**

`pdf-viewer.component.scss` has commented-out debug tints (`//background: rgba(255, 0, 128, 0.18)` and `//background: rgba(0, 255, 0, 0.3)`). The stale overlay block even says *Remove this block once caching behaviour is verified* — should be cleaned up before merge.

---

### :yellow_circle: Suggestions

**3. `MAX_ZOOM_LEVEL` jumped from 5 to 10** — is this intentional? At 10x with a large PDF, canvas memory consumption could be significant. Consider whether a lower ceiling (e.g. 6-7) would suffice.

**4. `getInitialZoom` clamps to `Math.min(1, fittingZoom)`** — tall pages in small viewports start zoomed out, but wide pages in narrow viewports don't get similar treatment. Worth considering depending on the document corpus.

**5. `border-radius` changed from `10px` to `0` on page skeletons** — minor visual change to the loading state. Was the rounded look intentional in the original design?

---

### :white_check_mark: What looks good

- **Handover service** is cleanly separated with clear lifecycle (`begin` → `releasePage` → `end`), safety timeout, and debug logging.
- **CSS-transform preview** during wheel zoom avoids expensive PDF.js re-renders during rapid scroll — correct pattern.
- **Text-layer rotation** with `data-main-rotation` CSS and the `correctTextLayerScaleX` fix for 90/270 degrees is correct.
- **Scroll anchoring** math for both zoom and rotation is thorough (preserving cursor position, horizontal ratio, page-relative offset).
- **OCR service change** correctly invalidates `pageDimensions` when OCR rescales pages.
- The `ZOOM_SETTLE_DELAY_MS` debounce prevents render thrashing during rapid wheel zoom.

---

**Verdict:** Architecture and approach are solid. Main actionable items: clean up debug artifacts (#2) and consider tracking/cancelling timers in `releasePage` (#1) to avoid edge cases with rapid zoom gestures.

vej-ananas (approved):

PR #1023, opened 2026-07-08
hotfix #1022: not all assets published

Dead0utside:
#1022


apps/sync-asset-sg/src/core/export-to-view.service.ts
```diff
@@ -45,7 +45,7 @@ interface AssetInfo {
   };
 }
 
-const BATCH_SIZE = 500;
```
vej-ananas, line 48: I would put it at 100 - we dont really know the current size and we probably dont want to keep retrying.

vej-ananas (approved):

PR #1034, opened 2026-07-31
fix: add explicit ES mapping for data field and use sourceFields for retrieval

uhon:
- Add 'data' field with 'index: false' to asset and file ES mappings to prevent wasteful text analysis of JSON payloads on reindex
- Switch AssetSearchService from storedFields (fields API) to sourceFields (_source filtering), consistent with FileSearchService
- Fix off-by-one in getDateTimeString: getUTCMonth() returns 0-11


vej-ananas (approved):
lgtm

PR #1037, opened 2026-08-04
Hotfix/assets-1022: not all assets published

Dead0utside:
Hopefully a robuster fix for #1022 

Prevent JavaScript heap exhaustion (FATAL ERROR: Reached heap limit)
during the view sync by streaming file export instead of loading and
transforming every file of an asset batch at once.

- Drop `files: true` from the main asset query and stop building a file
  array in preparePublishedData(); the source result, transformed copies
  and Prisma insert payload are no longer retained simultaneously.
- Add exportFiles(), called from the outer batch loop after exportAssets()
  so the asset/contact arrays are collectible before file rows are fetched.
- Filter publishable files in the database (Normal for normalFiles assets,
  Legal for legalFiles assets); non-publishable files never load their
  large JSON columns (fulltextContent, pageRangeClassifications,
  pageDimensions).
- Fetch ids first, then process in fixed chunks of FILE_CHUNK_SIZE (10):
  fetch -> transform -> insert (skipDuplicates) -> drop refs -> next chunk.
- Add process.memoryUsage()/V8 heap-limit diagnostics per chunk and per
  asset batch; no payload-size estimation, no global.gc(), no reliance on
  --max-old-space-size.

### DIFF
--- a/apps/sync-asset-sg/src/core/export-to-view.service.spec.ts
+++ b/apps/sync-asset-sg/src/core/export-to-view.service.spec.ts
@@ -58,7 +58,7 @@ function file(id: number, assetId: number, type: 'Normal' | 'Legal', overrides: 
     id,
     assetId,
     type,
-    size: 1n,
+    size: BigInt(1),
     fulltextContent: null,
     pageRangeClassifications: null,
     pageDimensions: null,
--- a/apps/sync-asset-sg/src/core/export-to-view.service.spec.ts
+++ b/apps/sync-asset-sg/src/core/export-to-view.service.spec.ts
@@ -1,0 +1,334 @@
+import { SyncConfig } from './config';
+import { ExportToViewService } from './export-to-view.service';
+
+// Suppress log output during tests
+jest.mock('./log', () => ({
+  log: jest.fn(),
+}));
+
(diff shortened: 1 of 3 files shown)


apps/sync-asset-sg/src/core/export-to-view.service.spec.ts
```diff
@@ -0,0 +1,334 @@
+import { SyncConfig } from './config';
+import { ExportToViewService } from './export-to-view.service';
+
+// Suppress log output during tests
+jest.mock('./log', () => ({
+  log: jest.fn(),
+}));
+
+// Keep this in sync with FILE_CHUNK_SIZE in export-to-view.service.ts.
+const FILE_CHUNK_SIZE = 10;
```
uhon: We should share this constant in `sync-asset-sg/src/core/config.ts` or create a new constants file in `sync-asset-sg`

uhon: and the chunk size seems rather small. With around 4000 assets this means many round-trips. I'd set it to at least 100 and check the memory stats tomorrow

uhon (changes requested):
Looks good. Please see my note on the batch size.

uhon (approved):

PR #1043, opened 2026-08-14
chore/Assets-997: update pdfjs to version 6

Dead0utside:
chore(pdf-viewer): update pdfjs-dist to version 6; Fix breaking changes

### DIFF
--- a/apps/client-asset-sg/project.json
+++ b/apps/client-asset-sg/project.json
@@ -26,7 +26,7 @@
           },
           {
             "glob": "pdf.worker.*",
-            "input": "node_modules/pdfjs-dist/build/",
+            "input": "node_modules/pdfjs-dist/legacy/build/",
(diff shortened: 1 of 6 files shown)


uhon (changes requested):
As already experienced during development of the pdf-viewer, there is an error with non-existing browser polyfills.
Version 6 (and also later versions in the 5.x series) do not contain the polyfill Promise.Try anymore and therefore fail.
This update breaks the PDF-Viewer and has potentially many other impacts on handling PDFs (clientside and serverside).
Even if we can fix the PDF-Viewer, a update to version 6 must be thoroughly tested with a bunch of different PDFs (both server-side and client-side)

Image

uhon (approved):
great, that the Promise.Try issues can be fixed by using the legacy build. code is fine!

Issue #1045, opened 2026-08-17
Investigate and handle Cypress cybersecurity outage impact


Dead0utside:
Cypress security incident notice: August 7th, 2026
We are writing to make you aware that we are investigating a security incident affecting a third party analytics service we use, Metabase. Our investigation is ongoing. We will share updates about the incident and our response as they become available.
What happened
On August 6, 2026, Cypress was notified that a previously unknown vulnerability in the vendor's cloud platform was exploited. The vendor published a public disclosure here: [Metabase security update](https://url9123.cypress.io/ls/click?upn=u001.3Yz-2BTUXDcTxzgVQ-2FtFHxxyeXiLw6mPl6nR3r4pQWsb9q5MBmhRFVwzZ-2B8stVmwFDOEDKzxTrpMIfY-2FFNvSw0Ww-3D-3DiOjW_M2-2Fja47ixAyW8uglAsVxDEO7eSsb2vptnI0LGAD3UsKWKuM4BVakvHhD6HoiLoxWCv48a-2Fec9ypGjkZez-2BiAL3XMOoTNNuLsEi3hu65XZLIxJu8n8kAjfJ58TiHw8uffcgfkLADFytQ9ELqs5dw653CklYPVNNwpDhS3tGCwylQ63ZOyCp9XqlkUMfqBmtVejQP6pm6QNh9DZr4ck8kVH-2B6niRVrndFmmcrTsSWOjgR33CB1StFk0-2BtkSobpy-2FiXl0upkX2GkHGDen2AHIXS3yLRqMeAFb-2BXAKUvt9A9nEstlcId55w7IrjbFprcr5Pg4HqgHr-2BueZqdUGk2Z87JbJjSJVMAS0213NqZBwhLbeLVx58Zfn2LncoyDvE9y5V8). Based on the vendor's representations, we understand that the vendor has since patched the vulnerability and invalidated all active sessions.
We began our response and investigation immediately after being notified by the vendor with the assistance of external cybersecurity experts. As mentioned above, our investigation is in the early stages, and we have found no evidence of ongoing unauthorized access to our systems. We are in the process of investigating the full scope and impact of the incident.
What this may mean for you
As a precaution, Cypress is in the process of revoking the authorization tokens used to sign in to Cypress Cloud. As a result, you may be asked to sign in again, and GitHub may send you a routine notification that access was revoked.
What was not involved
[...]

PR #1046, opened 2026-08-18
Fix/assets-867: Stabilize Cypress E2E tests, improve test infrastructure and harden CI secret handling

Dead0utside:
#867 
#1045

This PR stabilizes the Cypress E2E test suite and improves the overall test structure and maintainability.

Main changes include:
- replacing fixed cy.wait(...) calls with deterministic request- and state-based synchronization
- introducing reusable Cypress helpers for selectors, Material components, API requests and common page interactions
- replacing brittle structural/framework-specific selectors with stable data-testid attributes
- fixing session/authentication handling and token validation
- improving fixture setup and cleanup, including reliable restoration after mutating tests
- increasing Cypress timeouts and adding retries for CI runs
- fixing several tests whose previous assumptions no longer matched the actual application behavior
- improving classification tests so they reflect the current API state rather than stale fixture output
- removing unused Cypress component-test infrastructure and adding Jest tests for Cypress support logic
- fixing unnecessary DOM recreation in the asset file list by adding a stable trackBy
- moving the Cypress Cloud record key from the CLI argument to the CI environment to reduce unnecessary secret exposure

The main goal is to restore a consistently passing E2E suite while making the tests less dependent on timing, DOM implementation details and duplicated test logic.


uhon (approved):
Overall, it looks solid. Good that common helpers were separated and are reusable by future tests now.

Since it works and is a major improvement, i'll approve.

Informational:

Some Medium-severtiy issues AI review showed:

```
1.  deselectAllOptions  — infinite loop risk ( cypress/support/material.ts )
The  deselectNext  loop only terminates when  aria-selected="true"  options reach zero. Since the comment acknowledges elements are re-rendered/detached between clicks,  aria-selected  can be transiently stale, causing the same option to appear selected indefinitely. No max-iteration guard exists — this will silently hang until Cypress's overall timeout.

Fix: Track a max click count (= initially selected count) and throw a descriptive error if exceeded.

────────────────────

2. Intercept alias collision in  searchAndSelectAssetById  ( cypress/support/api.ts ,  cypress/support/pages/assetSearch.ts )
 interceptAsset  and  interceptAssetSearch  both use hardcoded aliases ( @asset ,  @assetSearch ). When  searchAndSelectAssetById  registers  @asset  before calling  searchAssetById  (which re-registers  @assetSearch ), the  @asset  intercept can fire during the search results render (before the click), causing  waitForSuccess  to match the wrong/already-consumed request.

Fix: Generate unique aliases per call (e.g., include the asset ID), or register  interceptAsset  only immediately before the click that triggers the asset load.

────────────────────

3. Timeout misconfiguration in  waitForEditorRoute  ( cypress/support/pages/assetSearch.ts )
 waitForEditorRoute  uses a 10,000 ms budget, but  defaultCommandTimeout  is set to 15,000 ms in this same PR. The very first  cy.location()  call can legally block for 15 s internally, exceeding the entire  waitForEditorRoute  timeout before it even starts polling — causing all retry attempts to fail immediately on slow CI.

Fix: Raise  timeout  to  20,000 + ms, or replace the manual polling loop with  cy.location('pathname').should('include', ...)  using an explicit timeout.

```

Nothing that crutial, but might be relevant if tests start to fail again in the future.
Can't decide whether it is worth fixing now, considering the limited budget.

Dead0utside:
> Nothing that crutial, but might be relevant if tests start to fail again in the future. Can't decide whether it is worth fixing now, considering the limited budget.

Thank you for your review. Since the budget is limited and the primary task was to make the tests pass again, I suggest making a note to come back to this PR in case tests start failing again. For now I'll just merge it

